PR #17, opened 2024-12-02
docs: foreground scope and source of submissions versus artifacts

cfm:
I'm happy to do more and/or put this elsewhere, but this seemed like the place to start.


cfm:
This is obsoleted by 87601fc717cb083e97e90ecc1754c6be22a5dd13.

PR #73, opened 2025-12-02
Draft webcat-infra-chain integration

lsd-cat:
### Steps

1. Load the validator set from disk.  *(TODO: Loaded from the block for now)*
2. Fetch the latest block from the CDN.
3. Verify the block using the validator set.
4. Check freshness: ensure the block date is greater than the last update's block date.
5. Fetch the `leaves.json` file.
6. Verify that the `app_hash` in `leaves.json` matches the `app_hash` from the block. (TODO)
7. Verify each leaf against the `app_hash`.  
8. Update the local database. (TODO)

TBD

### DIFF
--- a/extension/src/webcat/listeners.ts
+++ b/extension/src/webcat/listeners.ts
@@ -28,6 +28,10 @@ function cleanup(tabId: number) {
     }
     /* END */
     originState.current.references--;
+    // TODO: we should do this when an origin has 0 tabs AND gets evicted from the lru cache
+    // Theres no reason to do is as soon as all the tabs are closed, cause if it's a frequent website
+    // it will be re-added soon
+
     /* Here we could check if references are 0, and delete the origin object too */
     // TODO: if we do, we should also cleanup the listeners
     /*
--- a/extension/src/webcat/db.ts
+++ b/extension/src/webcat/db.ts
@@ -159,7 +159,7 @@ export class WebcatDatabase {
       }
     }
 
-    // 3. Update stored count (replace-only = exact count)
+    // 3. Update stored count
     await new Promise<void>((resolve, reject) => {
       tx.oncomplete = () => resolve();
       tx.onerror = () => reject(new Error("Transaction failed"));
@@ -171,9 +171,7 @@ export class WebcatDatabase {
   async getFQDNEnrollment(fqdn: string): Promise<Uint8Array> {
     const db = await this.ensureDBOpen();
 
-    //
     // 1. Positive-cache hit
-    //
     const originState = origins.get(fqdn);
     if (originState) {
       const cached = originState.current.enrollment_hash;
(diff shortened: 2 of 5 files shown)


extension/src/webcat/listeners.ts
```diff
@@ -24,7 +35,70 @@ async function updateList(db: IDBDatabase) {
...
+      { cache: "no-store" },
+    );
+    const block = await blockResponse.json();
+    console.log("[webcat] Update block fetched");
+    // 3 Verify block against validatorSet
+    const { proto: vset, cryptoIndex } = await importValidators(
+      block.validator_set as ValidatorJson,
+    );
+    const sh = importCommit(block as CommitJson);
+    const out = await verifyCommit(sh, vset, cryptoIndex);
```
redshiftzero: ✅ in `verifyCommit` we are verifying that we have 2/3+ voting power from valid signatures on precommits associated with the validator set 

Refs: https://github.com/freedomofpress/cometbft-ts/blob/main/src/lightclient.ts#L80-L214

extension/src/webcat/listeners.ts
```diff
@@ -24,7 +35,70 @@ async function updateList(db: IDBDatabase) {
...
+    );
+    const leaves = (await leavesResponse.json()) as WebcatLeavesFile;
+
+    // 6 Verify leaves file app_hash matches the block one
+    if (!arraysEqual(hexToUint8Array(leaves.proof.app_hash), out.appHash)) {
+      throw new Error("app hash mismatch");
+    }
+
+    // 7 Verify leaves against the canonical_root_hash and app_hash
+    const verifiedLeaves = await verifyWebcatProof(leaves);
```
redshiftzero: in `verifyWebcatProof`, we are verifying:
- ✅ the root of the reconstructed Merklized canonical state matches the canonical state root provided on the leaves endpoint

Refs: https://github.com/freedomofpress/ics23-ts/blob/main/src/webcat.ts#L209-L236

In `verifyCanonicalRootLink` (called by `verifyWebcatProof`) we are verifying:
- ✅ at least one Merkle proof is actually provided (via `if (proofBytes.length === 0)`)
- ✅ that we are using the _second_ proof provided by `cnidarium` which proves inclusion of the substore, and bailing if it does not exist for some reason
- ✅ that the canonical state root is included in the `AppHash` (via `verifyExistence`) - note I have not checked the `verifyExistence` logic in detail as I'm assuming this is out of the box ics23 merkle proof verification at this point

Refs: https://github.com/freedomofpress/ics23-ts/blob/main/src/webcat.ts#L184-L207

redshiftzero (approved):
awesome!

lsd-cat:
Filed https://github.com/freedomofpress/webcat/issues/77 and also we still have in TODO to hardcode the correct validatorset, but will do so when available.

PR #86, opened 2025-12-17
Allow updating extension while unlisted

redshiftzero:
unlisted browser extensions need to be submitted with a public `update.json` that we will periodically update when we do a new release. 

this is a temporary workflow as once the browser extension is listed this will become public. 

extension ID in our manifest is "webcat@freedom.press" hence the format below in `update.json` (i think this will change once mozilla signs)

R2 workflow is untested

### DIFF
--- a/.github/workflows/publish-update.json.yml
+++ b/.github/workflows/publish-update.json.yml
@@ -1,25 +1,19 @@
 name: Publish update.json to R2
 
 on:
-  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - 'update.json'
 
 jobs:
   publish-update:
-    runs-on: ubuntu-latest
-
-    steps:
-      - name: Checkout repository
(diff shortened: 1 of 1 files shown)


mig5:
After meeting with @redshiftzero, I've pushed a small change to the action, so that it uses our reusable actionslib workflow for R2.

You can see how it is very similar to that used on our packaging repos, e.g: https://github.com/freedomofpress/securedrop-yum-test/blob/main/.github/workflows/publish.yml

The secrets have been set in the repo for the bucket.

I'm pretty confident this will work, so I'm going to merge this in and see if it triggers as expected. Note the trigger tweak I made so that it should auto-run whenever a modification is made to `update.json` in the `main` branch.

mig5:
Aww dang, it didn't work: https://github.com/freedomofpress/webcat/actions/runs/20320155862/workflow

I'll push a hotfix soon.

Issue #110, opened 2026-02-17
Allow sha-xxx CSP to allow inline JS


lsd-cat:
As experienced, sadly some commonly used frameworks such as Next.js cannot work without inline JS. We currently disallow it due to how the WASM hooks works: they are injected at the network level in all the loaded scripts, but not in HTML pages.

Relevant for https://github.com/freedomofpress/webcat/issues/85

We could add a `<script>` tag dynamically each time we pass a `main_frame` or `sub_frame` at the network level. However, that would need to respect the page CSP, which means we'd have to dynamically compute the has of the hooks and add them to `script-src` as `sha-xxx`. However, this might be forbidden by AMO policies 

From https://extensionworkshop.com/documentation/publish/add-on-policies/:
 > Add-ons must not relax web page security headers, such as the Content Security Policy.

Though not much, I think we'd be technically 'relaxing' it by adding one item.

What we can also do is inject the hooks in any main_frame and sub_frame and hope for the best as content-scripts. However, `browser.tabs.executeScript` injects in an isolated "content_script" world, thus we'd need to inject a script that injects the hooks. I think this is even more of hacky solution than others, and the injected script would still have to follow the page CSP, having the same problem as before.

I would tentatively implement the first one, but it could be nice to get feedback by Mozilla to know if it's allowed.

lsd-cat:
Improved https://github.com/freedomofpress/webcat/pull/111 and merged

PR #111, opened 2026-02-17
feat: inject WASM hooks via content_scripts

lsd-cat:
Attempts to address https://github.com/freedomofpress/webcat/issues/110

In practice, Firefox content_script isolation can be bypassed by accessing `wrappedJSObject`. There's warning in the Mozilla dodcumentation documenting that that is possible and that it also breaks the isolation.

This PR refactor the WASM hooks to be able to run both via direct injection (via webRequest) and as a content_script. When injected as a content_script, there's a temporal guarantee that they will start executing before the rest of the page. Since they are injected (and the injection is awaited) after headers validation, but before forwarding them, we have a significant head start against any other content.

The hooks are bit ugly and should probably be slightly refactored for linting and better TS conformance. There's also need of some test cases to check that they don't easily fail.

### DIFF
--- a/extension/src/webcat/validators.ts
+++ b/extension/src/webcat/validators.ts
@@ -277,9 +277,7 @@ export async function validateCSP(
   }
 
   // Step 3: think about scripts
-  // Here allowing hash would break the WASM hooking; as we are no longer injecting
-  // Via a content_script, but rather at the network level on script files, having embedded
-  // JS in HTML page could break the assumptions.
+  // We can now allow inline verified scripts via sha-, see https://github.com/freedomofpress/webcat/pull/111
 
   await validateDirectiveList(
     directives.ScriptSrc,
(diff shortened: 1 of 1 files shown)


lsd-cat:
I've refactored the hooks build and injection system quite a bit:
 - The hooks are now split in different files, and there's one built for network injection and one built for content script injections, since they use different techniques to overwrite the WASM object
 - As a consequence, they have now their own vite config files for building
 - The function that generates the hooks (`getHooks`) as now a parameter to choose from the two

The `content_script` injection is tricky. Initially, I just did a `executeScript` in the target tab after header validation (which ensure that at that point WEBCAT metadata have been fetched and verified). However, the execution there is often (but not always too early): if the tab is not yet in an `OnCommitted` stage, then the injection will either fail silently or not be applied to the right context. As such, after header validation we just add a self-deleting listener for the tab, that triggers the injection as soon as the tab state changes to `OnCommitted`. This should be pretty reliable to inject before any actual content is executed, but not too early.

It's also important to remember when adding listeners that if the same function is added twice with different parameters, Firefox treat them as the same listener. Thus, we need to generated a function object each. This logic is suboptimal for performance, and could be switched for a global queue, but I like to keep things scoped this way for now.

Issue #132, opened 2026-03-18
(Almost) Drop the WASM hooks


lsd-cat:
Seeing the MDN description here:

  - https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/JavaScript_interface/instantiateStreaming_static

  > Note: For this to work, .wasm files should be returned with an application/wasm MIME type by the server.

 - https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/JavaScript_interface/compileStreaming_static

However, this page doesn't specify the same.

I wonder if we can drop support for the non streaming methods (compile/instantiale/Module) and just move verification at the network layer by checking the mime type of the response. More details later, we clearly need some experiments to see if the mime type is really a must.

lsd-cat:
We did not end up dropping the hooks, but they have been refactored as part of https://github.com/freedomofpress/webcat/pull/150. All the WASM methods are still supported.

Issue #131, opened 2026-03-12
NS_ERROR_FAILURE error when verifying webpage


sachaservan:
I am unable to have webcat verify the webpage. Running on MacBook Pro M1 with Firefox 148.0.2. 

<img width="876" height="298" alt="Image" src="https://github.com/user-attachments/assets/dbe2bd71-9464-438d-a2e2-94b2d028d7d8" />

# Error Logs: 

Uncaught (in promise) NS_ERROR_FAILURE: 
    onstop moz-extension://9326dd30-6251-407d-bfd2-720609aada5a/dist/bundle.js:13

Uncaught (in promise) NS_ERROR_FAILURE: 
    onstop moz-extension://9326dd30-6251-407d-bfd2-720609aada5a/dist/bundle.js:13
    Ml moz-extension://9326dd30-6251-407d-bfd2-720609aada5a/dist/bundle.js:13
    jo moz-extension://9326dd30-6251-407d-bfd2-720609aada5a/dist/bundle.js:13
    onBeforeRequest moz-extension://9326dd30-6251-407d-bfd2-720609aada5a/dist/bundle.js:13

Issue #129, opened 2026-03-11
WASM hooks are injected in the main_frame when only a sub_frame is loaded


lsd-cat:
As the title says, a non-enrolled site might load from an iframe an enrolled site. The hooks of the sub_frame were injected in the main_frame, breaking it. Even if that was not the case, there would be a mismatch with WASM auth. 

https://github.com/freedomofpress/webcat/commit/c7480970ede4000d4e9004d7c2d35624ece57cd4 is a temporary fix but we should test more cases

PR #139, opened 2026-03-30
Move WASM validation at the network-level

lsd-cat:
This PR is an attempt at implementing https://github.com/freedomofpress/webcat/issues/132. It also contains caching changes that could affect https://github.com/freedomofpress/webcat/issues/137 and a fix for https://github.com/freedomofpress/webcat/issues/131. It also partially affects https://github.com/freedomofpress/webcat/issues/129 since hooks here now do not contain per-origin information, and thus are generic.

As explained in the introdunction issue, the aim is to move the WASM validation logic at the network level. The MDN documentation states that `compileStreaming` and `instantiateStreaming` requires the `content-type` to be set to `application/wasm`. We can thus detect at the network level when that is the case, and enforce integrity.

Similarly, the hooks then just delete the non streaming methods: `compile`, `instantiate` and `Module`. The change also affects the manifest structure: since WASM is now validated as every other asset, there's no need for a dedicated `wasm` array, also providing better enforcement since validation has to be a dictionary match instead of a set match. However, to support _old_ manifests, there's still backward compatibility (which we could remove soon, since we are in alpha).

Clearly, this breaks web applications that use non-streaming methods. In my experience, those are rare, and it is also something that we can detect at manifest building time and warn or fix automatically. If we merge the changes here, we should also update  webcat-docs, webcat-spec, and webcat-cli.

Additionally, the response validation is moved from being called in `onBeforeRequest` to `onHeadersReceived`, because the `content-type` header is then needed during response validation. Apparently, that changes the cache visibility, since now even cached responses have a length. To address that, I'm now skipping re-validation of `304` response statuses instead of checking the length.

Lastly, hooks are now injected only via the content_script once, and at the network level only when `tabId` is < 0, trying to target only workers. 

Whether we merge this or not, some changes can be backported anyway, namely the fixes, possibly the cache detection, and the network injection only on workers.


lsd-cat:
I'm still architecturally uncertain on whether we should merge a corrected version of this patch. On one hand, I think simplifying the hooks is an improvement, especially by reducing the amount of network injection. On the other hand, it's another manifest change and we drop some functionality. I think we might end up changing the manifest anyway before a beta release, but still.

jupenur:
Getting rid of the complicated hooks and synchronous SHA-256 implementation seems like a pretty big improvement to me. The manifest change is nice, but not strictly necessary---IIUC we could keep backwards compatibility indefinitely if we wanted to. The only real counter-argument to merging this is then the reduced functionality. Maybe we need to get some numbers on how common the non-streaming WASM methods are before we make a decision?

jupenur:
Some numbers from PublicWWW, GitHub, and SourceGraph:

|WASM API|PublicWWW|GitHub|SourceGraph|
|---:|---:|---:|---:|
|compile|[51 sites](https://publicwww.com/websites/%22WebAssembly.compile%22/)|[8.7k files](https://github.com/search?q=%2FWebAssembly%5C.compile%5Cb%2F+%28language%3AJavaScript+OR+language%3ATypeScript%29&type=code)|[5963 hits in 1545 repos](https://sourcegraph.com/search?q=WebAssembly%5C.compile%5Cb+count:40000+%28lang:JavaScript+OR+lang:TypeScript%29+fork:yes+archived:yes&patternType=regexp&case=yes&sm=0)|
|instantiate|[7173 sites](https://publicwww.com/websites/%22WebAssembly.instantiate%22/)|[47.9k files](https://github.com/search?q=%2FWebAssembly%5C.instantiate%5Cb%2F+%28language%3AJavaScript+OR+language%3ATypeScript%29&type=code)|[22638 hits in 5112 repos](https://sourcegraph.com/search?q=WebAssembly%5C.instantiate%5Cb+count:40000+%28lang:JavaScript+OR+lang:TypeScript%29+fork:yes+archived:yes&patternType=regexp&case=yes&sm=0)|
|compileStreaming|[4247 sites](https://publicwww.com/websites/%22WebAssembly.compileStreaming%22/)|[2.7k files](https://github.com/search?q=%2FWebAssembly%5C.compileStreaming%5Cb%2F+%28language%3AJavaScript+OR+language%3ATypeScript%29&type=code)|[2153 hits in 583 repos](https://sourcegraph.com/search?q=WebAssembly%5C.compileStreaming%5Cb+count:40000+%28lang:JavaScript+OR+lang:TypeScript%29+fork:yes+archived:yes&patternType=regexp&case=yes&sm=0)|
|instantiateStreaming|[343 sites](https://publicwww.com/websites/%22WebAssembly.instantiateStreaming%22/)|[24.7k files](https://github.com/search?q=%2FWebAssembly%5C.instantiateStreaming%5Cb%2F+%28language%3AJavaScript+OR+language%3ATypeScript%29&type=code)|[20131 hits in 3486 repos](https://sourcegraph.com/search?q=WebAssembly%5C.instantiateStreaming%5Cb+count:40000+%28lang:JavaScript+OR+lang:TypeScript%29+fork:yes+archived:yes&patternType=regexp&case=yes&sm=0)|

On PublicWWW, `instantiate` has three apparent sources: two video streaming libraries used on a lot of adult sites, and [Flutter](https://flutter.dev/). Flutter is also a source for `compileStreaming`.

The hits on GitHub and SourceGraph could also contain Node.js software, not just stuff that runs in the browser. But in general it seems clear that the non-streaming `instantiate` API is significantly more common than any of the others.

lsd-cat:
Superseded by #150

PR #141, opened 2026-03-31
Fix integration tests

jupenur:
As mentioned in issue #136, the integration tests are broken in a bunch of ways. This PR attempts to fix some of those issues:

1. The testapp `webcat.config.json` now matches the current schema
2. The project-level Makefile can now be used to run integration tests: it also packages the test extension
3. The server for serving `/testing-list` is now a built-in fixture
4. `/testing-list` is automatically populated with the correct enrollment hash
5. The root path of the target app is now parameterized
6. The tests related to `x-sigstore-*` headers are now gone, as they no longer seem to be relevant and did not pass
7. The remaining tests look for specific error codes on the page instead of a generic integrity error
8. All tests pass

~~This PR still doesn't automatically generate the WEBCAT bundle for testapp; that's the only manual step remaining in running the integration tests.~~ The `webcat.json` file could also probably be deleted from the testapp but I didn't dare to do that yet in case it's needed for something.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 dist/
 extension/data/
+apps/*/.well-known/webcat/
--- a/apps/testapp/webcat.config.json
+++ b/apps/testapp/webcat.config.json
@@ -1,7 +1,8 @@
 {
-    "app_name": "Webcat Demo App",
-    "app_version": "0.1",
-    "comment": "https://github.com/freedomdofpress/webcat/apps/testapp",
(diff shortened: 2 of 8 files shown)


jupenur:
The Sigsum `demo.sh` is what I used for the manual step, but I think it would be nice to automate it---that way adding new test apps and making changes to existing ones becomes trivial. The non-determinism it introduces should not be an issue for most tests. For the ones that it is, we can hook the bundle and replace it with a static file.

lsd-cat (approved):

PR #144, opened 2026-04-06
WIP: Add more basic tests

lsd-cat:
As discusses elsewhere, integration tests are slow and it's pretty easy to test individual functions using `vitest` as we already do. This draft PRs aims to collect some more of these tests to add to the suite. In general these tests cases are also easier to AI generate.

I'm starting with tests for the OriginState class and will add more while PR is in draft. OriginState reminds me that there's also behavior that was never tested, such as the fallback manifest logic and the possibility of signaling manifest updates.

### DIFF
--- a/extension/tests/webcat/headers.test.ts
+++ b/extension/tests/webcat/headers.test.ts
@@ -1,0 +1,227 @@
+import { describe, expect, it, vi } from "vitest";
+
+import {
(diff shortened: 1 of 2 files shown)


lsd-cat:
Might still be useful later but not in this format, especially after refactoring.

PR #148, opened 2026-04-14
fix: cleanup code and do update check time-relative

### LINKED ISSUE #77: Improve and fix update process
Currently the update function always tries to update if there's a new block. However, there could be a new block but not a new list state and as such the canonical rot hash might be the same. Currently, this is hidden in https://github.com/freedomofpress/ics23-ts but the info should be made available and verified to the caller.

lsd-cat:
This PR removes some unused code. Furthermore, currently update logic was setting an interval at the end of which a list update would be attempted again. However, this behaved poorly with situations like sleep or suspending. In those cases, the interval would pause and resume after resume, thus not fetching an update even if many hours have passed.

The new logic checks every 5 minutes wether enough time (using the clock) has passed since the last update check, and in case triggers one. Thus, in case of suspension, only the 5 minute timeout has to be fully awaited. There's also associated tests.

Resolves https://github.com/freedomofpress/webcat/issues/77

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -16,6 +16,7 @@
   },
   "incognito": "spanning",
   "permissions": [
+    "alarms",
     "browsingData",
     "webNavigation",
     "webRequest",
--- a/extension/src/background.ts
+++ b/extension/src/background.ts
@@ -9,18 +9,33 @@ import {
 } from "./webcat/listeners";
 import { FRAME_TYPES } from "./webcat/resources";
 import { setErrorIcon } from "./webcat/ui";
-import { update } from "./webcat/update";
-import { initializeScheduledUpdates } from "./webcat/update";
+import {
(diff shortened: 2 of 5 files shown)


jupenur:
We should migrate to the [`alarms`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/alarms) API for future-proofing reasons, and this seems like a good PR to do it in. The `setTimeout` and `setInterval` methods are incompatible with manifest v3.

lsd-cat:
Thanks! I've moved it to the alarms API, and done some better error handling to avoid getting crashes on certain errors. There's also additional generated (but reviewed) tests, and the config variables have been moved to the config files.

I've checked the integration tests (so nice to have those!), and some manual testing.

jupenur (approved):
Nice! A lot easier to read, too.

PR #150, opened 2026-04-16
Hook WASM from content script

jupenur:
This PR removes the need to hook WASM from each JavaScript file individually. Instead it moves everything to the content script and installs hooks using `exportFunction`. Workers, SharedWorkers, and ServiceWorkers are still hooked by injecting directly to the script file. To detect dedicated Workers at the network level, the Worker constructor is also hooked.

This PR also fixes some bugs; most notably it adds the missing `compileOptions` parameter to all WASM methods that support it.

This is an alternative to PR #139 and only one of the two should be merged.

### DIFF
--- a/extension/src/globals.ts
+++ b/extension/src/globals.ts
@@ -1,9 +1,13 @@
 import { lru_cache_size, lru_set_size } from "./config";
 import { LRUCache, LRUSet } from "./webcat/cache";
 import { WebcatDatabase } from "./webcat/db";
+import { stringToUint8Array, Uint8ArrayToBase64Url } from "./webcat/encoding";
 import { OriginStateHolder } from "./webcat/interfaces/originstate";
 
 export const origins = new LRUCache<string, OriginStateHolder>(lru_set_size);
 export const nonOrigins = new LRUSet<string>(lru_cache_size);
 export const tabs: Map<number, string> = new Map();
 export const db = new WebcatDatabase("webcat");
+export const hookMarker = stringToUint8Array(
+  `__WEBCAT_HOOK__{${Uint8ArrayToBase64Url(crypto.getRandomValues(new Uint8Array(32)))}}\n`,
+);
--- a/extension/src/webcat/genhooks.ts
+++ b/extension/src/webcat/genhooks.ts
@@ -2,16 +2,18 @@ import contentHooks from "./../../dist/hooks/content.js?raw";
 import pageHooks from "./../../dist/hooks/page.js?raw";
 import { hooksType } from "./interfaces/base";
 
-export function getHooks(type: hooksType, wasm: string[], key: string) {
+export function getHooks(type: hooksType, wasm: string[]) {
   // This just patches the script string dynamically adding per-origin WASM hashes
   if (type === hooksType.page) {
-    return pageHooks
-      .replace('["__HASHES_PLACEHOLDER__"]', JSON.stringify(wasm))
-      .replace("__KEY_PLACEHOLDER__", key);
+    return pageHooks.replace(
+      '["__HASHES_PLACEHOLDER__"]',
+      JSON.stringify(wasm),
+    );
   } else if (type === hooksType.content_script) {
-    return contentHooks
-      .replace('["__HASHES_PLACEHOLDER__"]', JSON.stringify(wasm))
-      .replace("__KEY_PLACEHOLDER__", key);
+    return contentHooks.replace(
+      '["__HASHES_PLACEHOLDER__"]',
+      JSON.stringify(wasm),
+    );
   } else {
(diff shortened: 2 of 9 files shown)


jupenur:
I should add some integration tests that cover WASM handling, too, but I think that's best done in a separate PR. This PR has been tested manually against https://test.demo.webcat.tech/ and https://integrity-error.demo.webcat.tech/corrupted-wasm/.

lsd-cat:
Thanks! Given your findings about usage of the sync versus async WASM methods, I think this is likely a better solution than #139, especially since you managed to get it working exclusively through the content script (besides worker). I'll probably take a few days to do some additional testing due to the complexity of catching all edge cases, but I think this is the best direction for now :)

lsd-cat:
I've been testing it and it seems quite solid and reliable! I was brainstorming a bit if we had a way to avoid hooking `Worker`, even though we can't rely on the `tabId`. It seems that the `details` object of the response it's actually indistinguishable for any other script resource. However, the request is sent with `Sec-Fetch-Dest: worker`, thus we can detect it's being requested, though this probably would require matching requests with responses, and thus maybe keeping some state and too much complexity?

The other thing I was thinking, is that with the network based approach for WASM, which would work only for the streaming methods, there's the additional security benefit: it would be possible to match a precise WASM blob to the specific call, while now we have to to a check in the set (in theory, a server could swap blobs, though there's not a lot of exploitability). I don't think this is worth a hybrid and it  could also cause manifest building issues (ie: which blobs do you put in the set and which blobs in the file list?).

jupenur:
> However, the request is sent with `Sec-Fetch-Dest: worker`, thus we can detect it's being requested, though this probably would require matching requests with responses, and thus maybe keeping some state and too much complexity?

Good catch! I was looking for a way to detect Workers at the network level but completely missed this one. I'll see if I can adjust to make it work for us. I completely agree getting rid of the Worker hooking would be nice.

> The other thing I was thinking, is that with the network based approach for WASM, which would work only for the streaming methods, there's the additional security benefit: it would be possible to match a precise WASM blob to the specific call, while now we have to to a check in the set (in theory, a server could swap blobs, though there's not a lot of exploitability). I don't think this is worth a hybrid and it could also cause manifest building issues (ie: which blobs do you put in the set and which blobs in the file list?).

Even at the network level the matching of hashes to URLs is at best superficial: because we allow local redirects, a server could at any point switch from serving a script on `/scripts/foo.js` to redirecting it to `/scripts/bar.js`. WEBCAT would then validate responses against the latter URL even when the original request is for the former. I don't see a security benefit in tying hashes to URLs as long as redirects are allowed.

lsd-cat:
> 
> Even at the network level the matching of hashes to URLs is at best superficial: because we allow local redirects, a server could at any point switch from serving a script on `/scripts/foo.js` to redirecting it to `/scripts/bar.js`. WEBCAT would then validate responses against the latter URL even when the original request is for the former. I don't see a security benefit in tying hashes to URLs as long as redirects are allowed.

You are right, and when I wrote redirects I mostly thought about the main navigation, and I didn't realize the effect on subresources. We have the filenames in the manifest because we'd want to prevent the server from swapping resources (the classic example is a success indicator with an error indicator), so I think the fact that that's doable is a security issue. I'll file a related ticket

jupenur:
Worker detection now happens at the network level using another filter added from an onBeforeSendHeaders listener. The new filter only injects a marker, which the validation filter replaces with the real hooks. This is because the new filter runs first and injecting the hooks directly would mess up validation.

The new logic also detects AudioWorklets and PaintWorklets but I haven't tested that part. I'll create a separate ticket to test Worklets. PaintWorklets aren't supported in Firefox yet, but for future-proofing reasons it makes sense to cover that case already.

lsd-cat:
If we are passing state anyway, maybe out-of-bound is a bit cleaner than inbound, though then we'd have to manage the set where we track that, but that'd avoid the double filter and a call to crypto.subtle, I wonder what you think? Here's a patch, I didn't want to push to your branch :) 

https://github.com/freedomofpress/webcat/commit/513ef7f55649df6f03acd6c7bdf4cb1ef166e3ad

jupenur:
> If we are passing state anyway, maybe out-of-bound is a bit cleaner than inbound

That was my original plan as well, but then I started thinking ahead at the potential manifest v3 migration, after which we don't have reliable global state. I'm not sure what the exact v3 background service worker lifecycle is, but my concern is that a malicious server could serve response headers so slowly that the v3 worker gets shut off between onBeforeSendHeaders and onHeadersReceived: if we track the state in a global, it would get lost and the extension would fail open, which is a security issue.

lsd-cat:
Got it, let's merge and in case revisit strategies after a successful mv3 migration.

PR #153, opened 2026-04-21
Add WASM integration tests

jupenur:
This PR adds new integration test that cover WASM hooking and validation.

~~Due to minor differences in logging, none of the tests pass in main but do after #150 is applied. If for some reason we decide not to merge #150, the expected log entries are easy to adjust to match the main branch.~~


lsd-cat:
Tested and lgtm, thanks so much!

PR #156, opened 2026-04-21
Allow local redirects only on main frame navigation

### LINKED ISSUE #151: Redirects allow for resource swapping
As @jupenur pointed out it https://github.com/freedomofpress/webcat/pull/150#issuecomment-4286312021 the current redirect policies allow a server to redirect a subresource to another. Validation would happen on the target filename and resources, allowing a malicious webserver to swap resources. It's possible that for resources that are not active content, that could even mean that the server can redirect them to unverified ones (ie: for images, for instance).

I think that's an issue, and we should probably restrict redirects only on main navigation events, or match redirects against the original filename entry.

lsd-cat:
Fixes https://github.com/freedomofpress/webcat/issues/151 at the expense of losing most redirects. An alternative would be to match the destination resource against the origin url. However, I'd fear that could bring an amount of confusion due to the desyncing of filenames and their content.


lsd-cat:
For a similar logic, we should probably also allow fallback errors on main navigation events.

jupenur (approved):
The code looks good, but I feel like there may be cases where apps depend on the kinds of server-side redirects this blocks. We may end up having to relax this later. For now I can't think of a concrete reason not to merge this, though.

lsd-cat:
Thanks, I'll merge and if we run into problems we might try the other approach (verify the target against the origin hash).

Issue #154, opened 2026-04-21
Improve consistency of blocking behavior


jupenur:
Currently, most resources that fail an integrity check trigger an obvious error that's surfaced in both the page action icon and as the error page. WASM resources fail silently, as do ServiceWorkers and SharedWorkers, but for two different reasons: The WASM behavior is a design choice related to hooking, but no longer as relevant thanks to #150. The ServiceWorker and SharedWorker behavior is due to a bug in the error handling, where the filter that detects the integrity error tries to display the error page on tab -1.

The behavior should be made consistent, probably by making sure the error page is always shown. In the case of SharedWorkers and ServiceWorkers, we might want to show the error page in all same-origin tabs since there's no one specific tab that the error affects.

jupenur:
I'll start working on this by adding Worker integration tests on top of #153 and then fixing the ServiceWorker and SharedWorker bugs.

jupenur:
#157 covers SharedWorkers, ServiceWorkers, and AudioWorklets and triggers the error page on those integrity errors. #158 commit cda15f1 covers WASM hooked from content scripts where message-passing is easy.

The hard part is WASM hooked from Workers and especially Worklets. Workers are solvable: One solution could be pinging some magic URL from the WASM hook and catching that in an onBeforeRequest listener. I've prototyped with the idea, it works, and is relatively clean. I don't have a solution for AudioWorklet, because there the APIs are much more limited; in particular, there's no network access of any form.

Should we just leave WASM handling as-is, i.e. fail silently on integrity errors, or is the best-effort approach worth it even if it doesn't cover AudioWorklets? FWIW compiling WASM inside an AudioWorklet should be extremely rare; Worklets are an obscure feature to begin with, and the recommended approach to using WASM inside Worklets is to compile them on the main thread and then pass the compiled module to the Worklet for instantiation. WDYT @lsd-cat?

lsd-cat:
Thanks for the investigation, as you say AudioWorklets are pretty rare (I've never noticed one) and the combination with WASM and an error seems almost more like of a development problem than surfacing the error to the user (ie: whoever is implementing WEBCAT in such an app, might have an hard time detecting such errors before deployment). We could probably provide better tooling around that or even running an integration test with a custom target app (and then our logic could detect both page console errors and extension console errors).

I'd lean towards not implementing workarounds for now and leave it as it is, unless you feel otherwise.

jupenur:
[...]

PR #157, opened 2026-04-21
Show error page when blocking a ServiceWorker

jupenur:
Previously the errorpage function threw an error when it was called after blocking a ServiceWorker or a SharedWorker because tabId was set to -1. Now the function looks up all relevant tabs and shows the error page in each one. This is part of the fix to #154.

This PR builds on top of #153 and adds integration tests for corrupted workers.


jupenur:
The latest commit also fixes AudioWorklets and adds tests for them. The WASM hooks were not working in AudioWorklets due to lack of `crypto` APIs, `location` and `btoa` in the Worklet context.

lsd-cat:
I do really like the tab-scanning logic to avoid messaging or other workarounds! Thanks for this, I think it's a nice UI improvement

PR #161, opened 2026-04-23
CI for integration tests for firefox, firefox-esr and tbb

lsd-cat:
It would be nice to have the tests run automatically. Since the project is open source, we shouldn't have billing problems related to the long duration of the tests.

The tests are split per target, so that we can quickly see if there's failures what would be affected. As we add tests, we might want to try to parallelize to speed them up, though in my experiments that had some reliability issues.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -86,14 +86,6 @@ jobs:
           firefox --headless -CreateProfile "default" 2>/dev/null || true
           timeout 5 firefox --headless || true
 
-      - name: Build test extension
-        working-directory: extension
-        run: make package-test
-
-      - name: Install test dependencies
-        working-directory: test
-        run: make install
-
       - name: Run integration tests
         run: |
           make test TESTARGS="--addon ../dist/webcat-extension-test.zip \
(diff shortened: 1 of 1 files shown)


lsd-cat:
It seems that test should pass now. The only failure is related to a Github outage. It seems like if at least the platforms run in parallel, testing now takes ~10min. It could be nice to add automatic benchmarking, but not with the same frequency, so maybe on tags only so that we can notice and keep track of changes before pushing to AMO or elsewhere.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,101 @@
...
+        working-directory: test
+        run: make install
+
+      - name: Run integration tests
+        working-directory: test
+        run: |
+          .venv/bin/pytest -v tests.py \
+            --addon ../dist/webcat-extension-test.zip \
+            --headless \
+            -k "${{ matrix.test-filter }}"
```
jupenur: Nit: This could be done in one step by running `make test TESTARGS='--addon ../dist/webcat-extension-test.zip --headless -k "${{ matrix.test-filter }}"'` in the project root.

jupenur (approved):
Great stuff!

jupenur:
The project-level `make test` also runs `make package-test` and `make install` so now there's duplication. If we invoke pytest directly those two steps need to be run separately, but with `make test` they should be removed.

Issue #98, opened 2026-02-02
Tor Browser compatibility epic


lsd-cat:
This one to track issues and progress towards Tor Browser compatibility. In the current state, it does not work and has to be debugged.

lsd-cat:
So IndexedDB by default writes to disk because it could store large files and the risk of OOM is high.

See:
 - https://gitlab.torproject.org/tpo/applications/tor-browser/-/issues/26463
 - https://bugzilla.mozilla.org/show_bug.cgi?id=781982

It seems like Firefox now should have IndexeDb in private mode but not in the version used by tbb yet? Still digging in this

mikeperry-tor:
Also, here's uBlockOrigin struggling with this issue in Firefox some years ago: https://github.com/gorhill/uBlock/issues/3071.

PR #162, opened 2026-04-27
feat: add support for browser.storage.local backend

lsd-cat:
As suggested in https://github.com/freedomofpress/webcat/issues/98#issuecomment-4114271856 this PR adds support to `browser.storage.local` as a memory backend. The original class already supported testing IndexedDb and falling back on memory, while with this the order will be IndexedDb->storage.local->memory. It have tested in TBB and it seems to work.

Integration tests are passing. I think it's still useful to keep the memory fallback for now, but we might evaluate if to remove it in the future.

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -22,6 +22,7 @@
     "webRequest",
     "webRequestBlocking",
     "webRequestFilterResponse.serviceWorkerScript",
+    "storage",
     "tabs",
     "<all_urls>"
   ],
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -23,6 +23,7 @@
     "webRequestBlocking",
     "webRequestFilterResponse.serviceWorkerScript",
     "storage",
+    "unlimitedStorage",
     "tabs",
     "<all_urls>"
   ],
--- a/extension/src/webcat/db.ts
+++ b/extension/src/webcat/db.ts
@@ -1,232 +1,42 @@
 import { nonOrigins, origins } from "../globals"; // caching maps
-import { logger } from "./logger";
 import { extractHostname, extractRawHash } from "./parsers";
 
-type StorageMode = "indexeddb" | "storage.local" | "memory";
+const META_KEY = "block_meta";
 
-export interface ListMetadata {
-  hash: string;
-  treeHead: number;
+export interface BlockMeta {
+  blockTime: number;
+  rootHash: string;
 }
-
-interface WebcatStorageBackend {
-  settingsSet(key: string, value: number | string | bigint): Promise<void>;
-  settingsGet<T>(key: string): Promise<T | null>;
-  replaceList(leaves: readonly (readonly [string, string])[]): Promise<number>;
-  getEnrollmentByFqdn(fqdn: string): Promise<Uint8Array>;
-}
-
-// Settings keys
-const KEY_LAST_CHECKED = "lastChecked";
-const KEY_LAST_UPDATED = "lastUpdated";
-const KEY_ROOT_HASH = "rootHash";
-const KEY_LAST_BLOCK_TIME = "lastBlockTime";
-const KEY_LIST_COUNT = "listCount";
 
 export class WebcatDatabase {
-  private backendPromise: Promise<WebcatStorageBackend>;
-  public readonly storageMode: StorageMode;
-
-  constructor(private readonly name = "webcat") {
-    const { backendPromise, storageMode } = this.createStorageBackend(
-      this.name,
-    );
-    this.backendPromise = backendPromise;
-    this.storageMode = storageMode;
-  }
-
-  private async getBackend(): Promise<WebcatStorageBackend> {
-    return this.backendPromise;
-  }
-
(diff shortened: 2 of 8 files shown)


extension/src/webcat/db.ts
```diff
@@ -289,6 +344,77 @@ class IndexedDbStorageBackend implements WebcatStorageBackend {
...
+    // Remove old list entries
+    const all = await browser.storage.local.get(null);
+    const oldKeys = Object.keys(all).filter((k) =>
+      k.startsWith(`${this.prefix}list_`),
+    );
+    if (oldKeys.length > 0) {
+      await browser.storage.local.remove(oldKeys);
+    }
+
+    // Store new entries in chunks to avoid exceeding message size limits
```
jupenur: What's the limit mentioned here? I can't find anything on MDN.

extension/src/webcat/db.ts
```diff
@@ -289,6 +344,77 @@ class IndexedDbStorageBackend implements WebcatStorageBackend {
...
+  }
+
+  async replaceList(
+    leaves: readonly (readonly [string, string])[],
+  ): Promise<number> {
+    // Remove old list entries
+    const all = await browser.storage.local.get(null);
+    const oldKeys = Object.keys(all).filter((k) =>
+      k.startsWith(`${this.prefix}list_`),
+    );
```
jupenur: This seems a bit inefficient. Maybe we could maintain an index under another key instead and do something like

```typescript
const indexKey = `${this.prefix}index`;
const oldKeys = (await browser.storage.local.get(indexKey))[indexKey];
```

extension/manifest.json
```diff
@@ -22,6 +22,7 @@
     "webRequest",
     "webRequestBlocking",
     "webRequestFilterResponse.serviceWorkerScript",
+    "storage",
```
jupenur, line 25: Should we expect to run into the storage limits at any point when we scale? If yes, we might want to request the `unlimitedStorage` permission now rather than later.

extension/src/webcat/db.ts
```diff
@@ -43,30 +43,85 @@ export class WebcatDatabase {
...
+          reject(new Error("IndexedDB write probe transaction failed"));
+      } catch (e) {
+        reject(e);
+      }
+    });
+
+    return new IndexedDbStorageBackend(db);
+  }
+
+  /** Try to use browser.storage.local and verify it is writable. */
```
jupenur: Is there any scenario where the `browser.storage.local` API would exist but not be writable? The only one I can think of is if we hit the storage quota, but I'm not sure falling back to in-memory storage is the right solution.

jupenur (commented):
I'm wondering if we could just have `browser.storage.local` as the *only* storage backend? Is IndexedDB preferable for some reason?

lsd-cat:
> I'm wondering if we could just have `browser.storage.local` as the _only_ storage backend? Is IndexedDB preferable for some reason?

My original choice was for speed of writes and reads, especially as # of entries increase. I tried to find again supporting documentation for that but I couldn't. Ideally I agree it would be best to drop this abstraction and just use an API that is always available. I've run a quick generated benchmark script and this is the result (using clear() and writing everything in a single transaction for both):

image

Seems like list import is the only significant slowdown, though negligible at our current numbers.

Instead of the manual key removal that as you say is inefficient, we could also call [clear()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/clear) because I don't think we are storing anything else. You are right and there's no per-transaction/per-message limit, it's an hallucination as it works in a single call even with 100MBs of data.

I wonder what you think in light of the potential performance difference? I'm open to use only `storage.local` now and revisit later if we think performance becomes relevant again.

jupenur:
Are the numbers for the current logic where the `browser.storage.local` backed awaits for each each chunk separately? Collecting the promises and then doing `await Promise.all(writes)` would probably be significantly faster.

In any case, I don't feel like the numbers are that bad for the scale we expect for now. Simplifying would probably be the greater benefit.

lsd-cat:
> Are the numbers for the current logic where the `browser.storage.local` backed awaits for each each chunk separately? Collecting the promises and then doing `await Promise.all(writes)` would probably be significantly faster.
> 
The insert there was a single put with all the values, no chunking, but awaiting that. Anyway I agree, I'm gonna refactor with just that for now!

lsd-cat:
I've dropped all the backends in favor of `browser.storage.local`. The database is cleared with `clear()` which should be fast enough, and then set again in a single transaction, including some block metadata. I don't super like that FQDNs and metadata are mixed as keys, but there shouldn't be the possibility of somebody adding a key without a TLD to the list (in the original WEBCAT i assumed FQDNs would be hashes to to have a predictable list size, but we haven't done that in the chain and thus we do not want to hash all the keys prior to insertion here).

The other update information useful for scheduling update is migrated to `browser.storage.session`, in that way it never hits the disk but should be MV3 ready for later.

jupenur (approved):
Beautiful!

Image

PR #165, opened 2026-04-29
Fix NoScript compatibility

### LINKED ISSUE #160: Tor Browser Security Levels above Standard cause integrity errors
When the Security Level setting in Tor Browser is set to Safer or Safest, WEBCAT fails to verify websites. The first load of an enrolled website after browser start may succeed, but subsequent loads fail with either `ERR_WEBCAT_FILE_MISMATCH` or `ERR_WEBCAT_CSP_MISMATCH`.

jupenur:
The NoScript extension causes integrity errors in WEBCAT for two reasons: it adds custom CSP headers and patches worker scripts, making them no longer match the manifest. 

This PR deals with the first issue by ignoring all CSPs other than the first one. It is safe to do, because additional CSPs can only restrict the page further, not relax any previous rules. NoScript always only appends new CSPs, never modifies the original.

The second issue is slightly more complicated. This PR adds a new response filter that injects a marker tag. The response filter is installed in the onHeadersReceived listener, making it the last filter to be added; both WEBCAT and NoScript install other filters in the onBeforeRequest and onBeforeSendHeaders listeners. This order is important: the new filter runs last, meaning the injected tag is the last thing injected before receiving content from the network. The existing validation filter can then use the injected tag to detect which parts of the content are coming from extensions (everything before the tag) and which from the network (everything after the tag), ignore injected content, and compute the correct hash for the original content.

Fixes #160.

### DIFF
--- a/test/helpers.py
+++ b/test/helpers.py
@@ -208,7 +208,8 @@ class TorBrowser(Browser):
         if override_profiles_path == "":
             override_profiles_path = TorBrowser.get_profiles_path()
         additional_configs["network.proxy.allow_hijacking_localhost"] = False
-        additional_configs.update(TorBrowser.SecurityLevel._get_config(security_level))
+        if security_level != TorBrowser.SecurityLevel.Standard:
+            additional_configs.update(TorBrowser.SecurityLevel._get_config(security_level))
         super().__init__(override_tbb_path, override_profiles_path, additional_configs)
         if len(allowed_addons) > 0:
             with open(self.profile_path.joinpath("extension-preferences.json"), "r") as file:
--- a/test/helpers.py
+++ b/test/helpers.py
@@ -5,12 +5,22 @@ import queue
 import logging
 import subprocess
 import os
+import ssl
 import sys
+import tempfile
 import threading
 import http.server
 import socketserver
-from base64 import b64decode
+import hashlib
+import datetime
+import ipaddress
+from base64 import b64decode, b64encode
 from pathlib import Path
+
+from cryptography import x509
+from cryptography.x509.oid import NameOID
+from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives.asymmetric import rsa
 
 # --- Patch subprocess.Popen to discard Firefox output ---
 _original_popen = subprocess.Popen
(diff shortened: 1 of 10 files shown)


jupenur:
Oh, and this PR also adds some logic to enable integration testing at different Tor Browser Security Levels. No tests, though, just the scaffolding.

jupenur:
Yeah the integration tests won't work directly because the logs look a lot different at higher security levels. NoScript disables a bunch of features which means some of the log entries won't be there and some errors will be that aren't there otherwise.

lsd-cat:
Thanks for figuring out this, it's great!

Since the race with NoScript is a bit delicate, I did spend some time to work out some of the issues in order to have the already existing changes work, especially in light of upcoming interception changes. It took a bit of debugging, cause from test output it seemed that a) after setting a safer setting, then tests would fail on standard tbb too despite the profile changes and b) NoScript at safer blocks JS in any non secure origin (but it treats `.onion` as https).

The best solution would be for NoScript to treat 127.0.0.1 as secure (as I think browsers already do), but in the meantime there's now TLS support on localhost. I've also moved the skipping logic early so that instead of ignoring results, the marked tests do not run at all, and made the skipping logic incremental. Locally everything was passing (or skipped) so let's hope for that.

I'm gonna do one more manual logic and testing pass tomorrow, but everything look fine!

jupenur:
Awesome work with the tests. I added a variant of the "basic" test case that omits all WASM and Workers and works on TBB Safer level.

test/helpers.py
```diff
@@ -151,12 +202,14 @@ def get_profiles_path():
...
         return Path(os.path.dirname(TorBrowser.get_binary_path())).joinpath("TorBrowser/Data/Browser/")
 
-    def __init__(self, override_tbb_path="", override_profiles_path="", additional_configs={}, allowed_addons=[]):
+    def __init__(self, override_tbb_path="", override_profiles_path="", additional_configs={}, allowed_addons=[], security_level=4):
         if override_tbb_path == "":
             override_tbb_path = TorBrowser.get_binary_path()
         if override_profiles_path == "":
             override_profiles_path = TorBrowser.get_profiles_path()
         additional_configs["network.proxy.allow_hijacking_localhost"] = False
+        if security_level != TorBrowser.SecurityLevel.Standard:
```
jupenur, line 211: Explicitly not setting the security level here if it's Standard, because TBB recently changed the Standard config. Leaving it at the default means the Standard config is always correct regardless of TBB version.

lsd-cat:
Thanks, did some more testing and it seems to work great :)

PR #166, opened 2026-04-30
WIP: Implement Tor circuit hints in BundleFetcher

jupenur:
With this change, BundleFetchers send a custom header that the Tor Browser can use to select an appropriate circuit for the request. The header is only sent on Tor Browser.

### DIFF
--- a/extension/src/webcat/listeners.ts
+++ b/extension/src/webcat/listeners.ts
@@ -3,6 +3,7 @@ import { db, origins, tabs } from "../globals";
 import { getHooks } from "./genhooks";
 import { hooksType, metadataRequestSource } from "./interfaces/base";
 import { WebcatError } from "./interfaces/errors";
+import { BundleFetcher } from "./interfaces/originstate";
 import { logger } from "./logger";
 import { validateOrigin } from "./request";
 import { FRAME_TYPES } from "./resources";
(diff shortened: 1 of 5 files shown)


extension/src/webcat/listeners.ts
```diff
@@ -263,3 +266,32 @@ export async function requestListener(
...
+export async function torCircuitListener(
+  details: browser.webRequest._OnBeforeSendHeadersDetails,
+): Promise<browser.webRequest.BlockingResponse> {
+  if (!details.requestHeaders) {
+    console.error("FATAL: request headers not available");
+    return { cancel: true };
+  }
+  if (isExtensionRequest(details)) {
+    let headers: browser.webRequest.HttpHeaders;
+    if ((await browser.runtime.getBrowserInfo()).vendor === "Tor Project") {
```
jupenur, line 279: This limits sending the circuit hints to TBB; on Firefox, no hint will be sent. Before merging, we may want to limit it further to specific versions of TBB, because until TBB implements its part, there's a risk of leaking information across origins.

lsd-cat:
Updates here https://gitlab.torproject.org/tpo/applications/tor-browser/-/work_items/44927#note_3434335

TLDR: I did propose two approaches: one use the header as implemented here, one adds a custom option to the `fetch` itself. I think the latter is cleaner, and more coherent with the platform API (and also smaller). We have to do some additional analysis described in the ticket, so the progress is partially blocked on that.

jupenur:
Closing this one since the preferred approach changed to using a dedicated API instead of a custom header.

PR #167, opened 2026-05-05
Add a test for a bypass attack with multiple tabs

jupenur:
Add a test for an attack that allows bypassing integrity checks: inducing an error in one tab while another tab is still loading causes the `webRequest` listeners for that origin to be unregistered, allowing responses to pass through unvalidated. The test FAILS on Firefox because there is an attack, but PASSES on TBB because TBB seems to enforce a strange ordering of the requests. The pass on TBB doesn't mean the issue couldn't be exploited on TBB under any circumstances.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -23,6 +23,7 @@ _tbb_safest_skips = {
     "corrupted_worker_test": "JavaScript fully disabled at this security level",
     "corrupted_sharedworker_test": "JavaScript fully disabled at this security level",
     "corrupted_audioworklet_test": "JavaScript fully disabled at this security level",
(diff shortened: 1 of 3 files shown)


lsd-cat:
Oh nice find! I'd lean towards merging, and rebasing https://github.com/freedomofpress/webcat/pull/168 to have it there, where I think it should pass.

jupenur:
I could also cherry-pick this directly to your branch on #168 so we don't have to merge tests that don't pass. Either way sounds good to me.

lsd-cat:
Let's merge, we should get it passing anyway!

Issue #133, opened 2026-03-19
Define frames/resources integrity logic


lsd-cat:
We have to decide when to verify resources and when to consider something verified. For instance, a non enrolled domain might load an iframe from an enrolled domain, such as the case with https://chat.tinfoil.sh. In that case, currently, the sub_frame is verified, and a "neutral" WEBCAT icon is shown ("WEBCAT is running"). In order to support third party resources from other enrolled domains, (ie: script-src webcat-origin.com) we must also verify and load webcat-origin.com. That is a MUST when it is loaded from an enrolled domain.

However, if non-enrolled.com loads scripts from enrolled.com but never loads a frame from it, due to the current interception logic, that assets doesn't get verified. Reasons for that are mostly performance: we currently intercept only when navigation is to a main_frame or a sub_frame event. We should probably always verify everything from enrolled domains, even if we don't display any indicator. We can probably think of dynamically adding and removing from interception lists, but we might want to have benchmarks for that.

lsd-cat:
I think `<embed>` count as object? And testing it seems like `<object>` can also load HTML iframe-like. I think currently we block `object-src` at the CSP level, but there's probably not a good reason to be that strict. I tend to agree that we should always verify anything from an enrolled domain, though that will change some of the interception logic again.

jupenur:
> I think currently we block `object-src` at the CSP level

Right, but that only helps with enrolled origins embedding stuff. The problem case here is a non-enrolled origin embedding an enrolled one: because embedding doesn't count as frame navigation, the tab doesn't get enrolled and the object response (from an enrolled origin, in a non-enrolled tab) doesn't get filtered, the embed loads and renders, and can compromise the locally persisted state of the enrolled origin.

lsd-cat:
I think after internal discussions we are set that we should always validate resources, no matter the frame navigation.

We have now two implementation choices:
 - We can return to the previous status, where we intercept `*` and then have a fast exit for non-enrolled domains (there's caching on top of IndexedDB so the general case should be fast. This in theory, scales "forever"
 - Otherwise, at startup and at list updates, we can dynamically extract the enrollment list and add a single listener that covers all of them. Performance-wise this is the best option: we do not even intercept in the case of non-enrollment. However, this likely will work well until we are in the low thousands of domains. I don't think that would be a problem anytime soon though.

I'm gonna try to implement the second option and open a draft PR.

jupenur:
[...]

PR #168, opened 2026-05-05
Always intercept enrolled domains

lsd-cat:
Attempt at fixing https://github.com/freedomofpress/webcat/issues/133. As described in the ticket, this approach adds a global listener, loading all the enrolled domains from the database. To my understanding, it should be ok until a few thousands enrollment, at which we should re-evaluate. It also wouldn't be reusable for Onion domains, as we wouldn't know the plaintext of them.

I've been delayed a bit because I am still getting occasional test failures which root cause I haven't fully understood.

### DIFF
--- a/extension/src/webcat/listeners.ts
+++ b/extension/src/webcat/listeners.ts
@@ -1,5 +1,6 @@
 import { endpoint } from "../config";
 import { db, origins, tabs } from "../globals";
+import type { WebcatDatabase } from "./db";
 import { getHooks } from "./genhooks";
 import { hooksType, metadataRequestSource } from "./interfaces/base";
 import { WebcatError } from "./interfaces/errors";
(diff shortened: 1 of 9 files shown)


extension/src/webcat/listeners.ts
```diff
@@ -266,3 +267,106 @@ export async function requestListener(
...
+  const urls = buildUrlPatterns(fqdns);
+
+  // The registration needs to be different from the existing one
+  const before = (details: browser.webRequest._OnBeforeRequestDetails) =>
+    requestListener(details);
+  const beforeHeaders = (
+    details: browser.webRequest._OnBeforeSendHeadersDetails,
+  ) => beforeHeadersListener(details);
+  const headers = (details: browser.webRequest._OnHeadersReceivedDetails) =>
+    headersListener(details);
```
jupenur, line 346: If, instead of wrapping in a closure every time, we registered the original function as a listener directly, wouldn't that remove the need to explicitly do `removeListeners` and otherwise be equivalent with the current code? It would also be more compatible with manifest v3, because then we'd no longer have to keep track of `currentListeners` across backrground service worker restarts.

lsd-cat: My understanding is that, if I do that, I'd need to remove the old one before adding the new one, and would leave a (really small) window with no listeners during the update time?

jupenur: Yeah I tried it out and you're right. I was expecting it to replace the filters if matching listener exists, but it looks like the addListener call is just a no-op in that case.

jupenur (approved):
I filed two new issues, #169 and #170, that are still problems after this is merged, but otherwise everything looks good. Did some basic manual testing in Firefox, too, and found nothing to complain about.

Great work! I think we're good to merge this.

lsd-cat:
Great! Thanks for the quick review :)

Issue #170, opened 2026-05-05
Origin might not be populated in getVerifiedManifestState


jupenur:
If the same enrolled site is loaded in two tabs simultaneously, and one triggers an error, the origin cache might get cleared. If the timing is right (or wrong), when the remaining tab triggers a call to getVerifiedManifestState, it might throw because the origin is not populated.

lsd-cat:
By looking a bit into this: pruning origins in the case of integrity errors is clearly not optimal. I think my logic when I wrote that is to have the case where the server refreshed assets or uploaded a new manifest, and the client got desynced (tgough in theory this can be managed via the webcat update header). The idea was to recover safely at the next refresh. Even if we didn't, I think there's still the case where a manifest validation error happens, like a timeout out fetch, an expired manifest, the wrong signature, etc. In that case we'd still need to prune, but we might risk the same condition?

I wonder if we should just apply the cache context isolation that we have to do for tbb globally, and then avoid this problems about potential concurrency.

jupenur:
> I wonder if we should just apply the cache context isolation that we have to do for tbb globally

You mean give each tab its own cache regardless of first-party origin? Or implement the TBB-style first-party cache isolation in all browsers? The former would solve it, I think, and #157 already solves the tricky case where tabs share a cache via a ServiceWorker.

Another option is to keep the pruning logic as it is and fix these points where we assume an origin exists, so that the non-existing origin is handled gracefully.

Yet another option would be to mark origins stale instead of pruning them, and then do special-case handling for stale origins where needed.

lsd-cat:
You are right. Maybe a simpler fix without touching the caching logic too much now could be to just populate the cache on success? We'd lose some performance since in the case of multiple concurrent tab from a cold start they could all do the fecthing+verification, but it would simplify the reasoning about timing quite a bit.

jupenur:
That sounds pretty acceptable to me. Multiple same-origin tabs loading concurrently from a cold start should be a rare enough occurrence for the practical performance hit to be minimal. Like, I can't imagine anyone having the same SecureDrop open in 12 different tabs and then restoring them in the next browser session.

PR #173, opened 2026-05-08
Fix content script race

### LINKED ISSUE #169: Race in content script injection
There's a race in the  content script injection logic:

https://github.com/freedomofpress/webcat/blob/80ec25a95929414062182d02e06f380cf086a1ae/extension/src/webcat/listeners.ts#L154-L169

The `onCommitted` event is non-blocking, so the `executeScript` call might come in late. We should be using the `scripting.registerContentScripts` API instead.

jupenur:
Fixes #169 by injecting the content script twice: first reliably, and then using the racy method. The first injection installs the hooks and blocks the async WASM methods until hashes become available. The second injection checks that the hooks are in place, updates the hashes, and unblocks WASM.

There is one failure mode that this PR doesn't solve: inline scripts calling the synchronous WASM methods before hashes are available. When that happens, the calls fail with an integrity error. I don't know of a way to make the hashes available synchronously without hacky synchronous XHR, which I want to avoid.

The PR includes an integration test that reliably wins the race against the current content script injection method but passes after the fix.


jupenur:
The remaining test failures are due to #170.

extension/src/webcat/listeners.ts
```diff
@@ -366,6 +366,29 @@ export async function installEnrolledListeners(
...
+  await browser.scripting.registerContentScripts(
+    fqdns
+      .filter((fqdn) => {
+        return !scriptIds.includes(fqdn);
+      })
+      .map((fqdn) => {
+        return {
+          id: fqdn,
+          js: ["dist/hooks/content.js"],
+          matches: buildUrlPatterns([fqdn]),
```
lsd-cat, line 427: I was reading the doc and I think we should add `matchOriginAsFallback`, as I think we could allow iframes from blob: or data: on the same origin, though that's still tricky.

extension/src/webcat/listeners.ts
```diff
@@ -366,6 +366,29 @@ export async function installEnrolledListeners(
...
+          runAt: "document_start",
+        };
+      }),
+  );
+  // Remove the content scripts whose fqdn is no longer enrolled
+  await browser.scripting.unregisterContentScripts({
```
lsd-cat, line 435: Earlier in the file there's the code to drop the network listeners in case the list is empty, and then a return which would make it so we never unregister the scripts. Honestly, I think we can just drop the earlier code, as we are maintaining the list and we do not expect it to reach zero :)

lsd-cat (commented):
I think we already had an issue, by looking at our policies, where I think it could be possible to have an iframe via `srcdoc` where the hooks don't get injected, and from there I think Wasm modules are also clonable and sendable via `postMessage`? I think `matchOriginAsFallback` would likely address this, and I also think this PR fixes https://github.com/freedomofpress/webcat/issues/129 but I'm curious of your opinion

extension/src/webcat/listeners.ts
jupenur: Good catch. Fixed.

jupenur:
> I think we already had an issue, by looking at our policies, where I think it could be possible to have an iframe via `srcdoc` where the hooks don't get injected, and from there I think Wasm modules are also clonable and sendable via `postMessage`?

The first part is clear (hooks not getting injected in `srcdoc`, `data:`, and `blob:`) and I'll try to add a test for that and then fix it. The second part I'm not sure about: yes, WASM modules can probably be cloned and sent via `postMessage`, but is that an issue? If it is, wouldn't it also be an issue for cross-origin frames? An app could compile a WASM module in a non-enrolled origin and then post it to an enrolled origin for instantiation, but there's nothing we can do about that. 

> I also think this PR fixes #129 but I'm curious of your opinion

I was actually already digging into this one and trying to write an integration test for it, but I can't for the life of me figure out how to access subframes via geckordp, so I didn't get very far.

jupenur:
It was slightly more complicated than just adding `matchOriginAsFallback` but I think I now have the `srcdoc` case covered. I also have an approach for testing #129 and will do that next before re-requesting review.

extension/src/webcat/listeners.ts
```diff
@@ -167,7 +167,7 @@ export async function headersListener(
       });
     };
 
-    browser.webNavigation.onCommitted.addListener(listener);
+    browser.webNavigation.onDOMContentLoaded.addListener(listener);
```
jupenur, line 194: Turns out onCommitted can be too early as well as late, so this had to be moved to onDOMContentLoaded. There is a performance impact because the WASM hashes will be delivered to the content script slightly later, but in any apps written in a reasonable way it should be minimal.

jupenur:
Tests should be passing now. There are a lot more of them, because this PR now adds an `in_frame` mode to `test_webcat`. The new tests did catch a couple of minor bugs, though, that are now fixed along with #129. Ready for review.

jupenur:
There's some flakiness that's now exacerbated by the increased number of tests. For now, my assumption is that it's flakiness related to the tests themselves, not flakiness in the behavior of WEBCAT, but in my next PR I'll work on getting better visibility to what's actually failing and why.

lsd-cat:
Looks good to me and thanks, this is so much work! I'm glad we also have the iframe testing now, it was really painful to do manually and I don't think i'd ever explored these many cases.

lsd-cat (approved):

Issue #171, opened 2026-05-06
Test webapp update scenario


lsd-cat:
There's code for the server to signal and trigger a manifest reload via `x-webcat-version`, however I've never tested that even manually. We should probably test (or change logic and improve it) as servers should be able to trigger manifests refresh when necessary. WAICT adopts a different approach: the manifest url is specified in a header, and they assumed a manifest at a url never changes. When a manifest is at a new path, that becomes valid. I think in WEBCAT we are fine with the fixed paths, which also makes enrollment and auditing simpler to reason about.

lsd-cat:
Automated testing implemented in https://github.com/freedomofpress/webcat/pull/179, we should eventually write docs but marking as resolved for now.

PR #174, opened 2026-05-11
Fix cross-tab race of origin validation

lsd-cat:
Attempt at fixing https://github.com/freedomofpress/webcat/issues/170. Create temporary origins object in a new global object, `pendingOrigins`, keyd per request id, which should be unique. Upon successful validation, promote to `origins` only if a newer version of the currently cached one.

### DIFF
--- a/extension/src/webcat/request.ts
+++ b/extension/src/webcat/request.ts
@@ -1,4 +1,4 @@
-import { origins, tabs } from "./../globals";
+import { origins, pendingOrigins, tabs } from "./../globals";
 import { db } from "./../globals";
 import { metadataRequestSource } from "./interfaces/base";
 import { WebcatError, WebcatErrorCode } from "./interfaces/errors";
@@ -16,6 +16,7 @@ export async function validateOrigin(
   url: string,
   tabId: number,
   type: metadataRequestSource,
+  requestId: string,
 ) {
   const enrollment_hash = await db.getFQDNEnrollment(fqdn);
   if (enrollment_hash.length === 0) {
@@ -47,10 +48,7 @@ export async function validateOrigin(
     tabs.set(tabId, fqdn);
   }
 
-  // If origin metadata are already loaded, just skip doing it again and return early
-  const originStateHolder = origins.get(fqdn);
-  if (originStateHolder) {
-    // Since we use cached info, we should still populate the popup with the cached info
+  if (origins.get(fqdn)) {
     return;
   }
 
@@ -74,7 +72,7 @@ export async function validateOrigin(
     enrollment_hash,
   );
   const origin = new OriginStateHolder(newOriginState);
-  origins.set(fqdn, origin);
+  pendingOrigins.set(requestId, origin);
 
   // See https://github.com/freedomofpress/webcat/issues/95
   await origin.current.fetcher.awaitAll();
--- a/extension/src/webcat/request.ts
+++ b/extension/src/webcat/request.ts
@@ -48,7 +48,10 @@ export async function validateOrigin(
     tabs.set(tabId, fqdn);
   }
 
-  if (origins.get(fqdn)) {
+  const cached = origins.get(fqdn);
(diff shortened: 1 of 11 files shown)


lsd-cat:
Seems like tests are passing. This PR adds some consideration for the scenario in https://github.com/freedomofpress/webcat/issues/171 but doesn't test it, and I feel like the update logic is probably still incomplete, but I'd leave it to a future issue as we work towards that.

extension/src/webcat/request.ts
```diff
@@ -47,10 +48,7 @@ export async function validateOrigin(
     tabs.set(tabId, fqdn);
   }
 
-  // If origin metadata are already loaded, just skip doing it again and return early
-  const originStateHolder = origins.get(fqdn);
-  if (originStateHolder) {
-    // Since we use cached info, we should still populate the popup with the cached info
+  if (origins.get(fqdn)) {
```
jupenur: I guess we want `get` here instead of `has` in order to update the LRU cache? Maybe add a comment to that effect because it isn't obvious otherwise.

jupenur (commented):
It seems to me like there's still a (tiny, but real) window between the call to `commitVerifiedOrigin` in `headersListener` and the call to `getVerifiedManifestState` in the response filter. During that window, the origin might still get purged from the cache, and I suspect an actively malicious server could actually still exploit it by carefully serving only headers, pausing before the response body, and at the same time triggering cache eviction by flooding it with unrelated entries.

Is there a way we could close that window?

jupenur:
> I suspect an actively malicious server could actually still exploit it by carefully serving only headers, pausing before the response body, and at the same time triggering cache eviction by flooding it with unrelated entries.

I was able to confirm this with an integration test and opened PR #175. Depending on how we choose to fix the issue, we may or may not want to merge the test, but at least it's there to help with fix development.

lsd-cat:
> I was able to confirm this with an integration test and opened PR #175. Depending on how we choose to fix the issue, we may or may not want to merge the test, but at least it's there to help with fix development.

Thank you, I think keeping the test is useful for future changes or regressions! I'm gonna work on addressing this today

lsd-cat:
Turns out https://github.com/freedomofpress/webcat/pull/168 was incomplete: it still called `validateOrigin` only on `main_frame` or `sub_frame` navigation. Thus, with the new logic, even without the race condition, loading an enrolled subresource without a cache hit would be a problem. My understanding is the current test failure is similar to that: when loading a subresources, the cache does the pruning, and the request fails. I think there's a bit more refactoring to do, mostly to unify the frame/non frame validation path, while keeping some conditional features (such as setting the status icon). I'm marking it back to draft until I address everything

lsd-cat:
Ok, it grew in scope more than I hoped, but here we are. Fixes #170 and the same restructuring also extends WEBCAT validation to sub-resources fetched by non-enrolled main frames, and commits the test coverage for the LRU cache issue (thanks Juho) and an additional test to cover the unenrolled loads enrolled subresource case. I think also the logic around signaling manifest updates has improved, but it's still untested (will follop up in https://github.com/freedomofpress/webcat/issues/171 eventually).

The main goal to eliminate the various races is to use the cache once: at request firing we pull from cache or create an origin from scratch and keep a reference to the object for the lifetime of the request. This approach doesn't need per tab state, but rather per-request state, though there's listeners to ensure to clean it up. It also now passes the previous tests that were failing. With this, I think we can also probably eliminate some of the TODOs/development guards that I had around the code to try to catch this type of races, which clearly didn't work

jupenur (approved):
This looks great. Pinning to the `pendingOrigins` map is a really nice solution. There's going to be some manual work in merging #173 after this, but the sooner we get this one in, the sooner I can get started on it 🐱

lsd-cat:
Thanks for the fast review and the brainstorming you are doing to find all the races! I'm going to merge it then, and sorry to interfere with #173 which such a big refactor

PR #177, opened 2026-05-15
Improve integration test organization

jupenur:
Two QoL improvements to integration tests:

1. Better error messages by using `pytest_check` instead of `assert` and hence checking all logs instead of bailing out on the first mismatch
2. More maintainable test organization by moving test IDs from the separate `ids` lists to `pytest.param` so you don't have to look up the ID and param separately.


lsd-cat:
Thanks, I think the logs are going to help significantly :)

Issue #176, opened 2026-05-13
Ensure we fail closed


lsd-cat:
Problems like https://github.com/freedomofpress/webcat/issues/170 have shown that in case of programming errors in our side, the default behavior is to fall through and allow unverified resources through. I think the ideal behavior, is that in case of missing objects, undefined calls or any other error that we haven't found yet, the system should fail blocking (only for enrolled domains) and tell the user to file a bug. I had some "guards" scattered around to try to catch these conditions, but they have clearly failed. Also, throwing doesn't block execution, thus when we through (or the system does) we should take active steps to cancel the ongoing request.

lsd-cat:
Finally done in https://github.com/freedomofpress/webcat/pull/178

PR #178, opened 2026-05-15
Fail closed during interception

lsd-cat:
Attempts to implement fixes for https://github.com/freedomofpress/webcat/issues/176. Basically, while I hope we've fixed the race conditions, the underlying problems is also that on throws the default behavior is to pass the request or content through. I think we should rather always drop in case of WEBCAT errors. In that way, if someone defies our logic, they are stuck with a DoS instead of a bypass.

The first commit here implements this only for the streamFilter, while similar behavior should be applied to the request interceptor and the headers validator. The alternative to one big try catch is a wrapper function, but I find that less readable. Curious if that looks reasonable to you @jupenur and in case I'll improve coverage before being ready.

### DIFF
--- a/extension/src/mocks/utils.mock.ts
+++ b/extension/src/mocks/utils.mock.ts
@@ -1,0 +1,39 @@
+export * from "../webcat/utils";
+
+import { SHA256 as realSHA256 } from "../webcat/utils";
+
+// Overrides SHA256 function with a throwing one when the test marker
+// is found
+const FAIL_CLOSED_TRIGGER = new TextEncoder().encode(
+  "__WEBCAT_TEST_FAIL_CLOSED__",
+);
+
+function containsTrigger(view: Uint8Array): boolean {
+  outer: for (let i = 0; i <= view.length - FAIL_CLOSED_TRIGGER.length; i++) {
+    for (let j = 0; j < FAIL_CLOSED_TRIGGER.length; j++) {
+      if (view[i + j] !== FAIL_CLOSED_TRIGGER[j]) continue outer;
+    }
+    return true;
+  }
+  return false;
+}
+
+export async function SHA256(
+  data: ArrayBuffer | Uint8Array | string,
+): Promise<ArrayBuffer> {
+  let view: Uint8Array;
+  if (typeof data === "string") {
+    view = new TextEncoder().encode(data);
+  } else if (data instanceof Uint8Array) {
+    view = data;
+  } else {
+    view = new Uint8Array(data);
+  }
+  if (containsTrigger(view)) {
+    console.log(
+      "[TESTING] utils mock SHA256 — throwing to exercise failClosed",
+    );
+    throw new Error("[TESTING] fail-closed trigger");
+  }
+  return realSHA256(data);
+}
--- a/extension/src/webcat/interfaces/errors.ts
+++ b/extension/src/webcat/interfaces/errors.ts
@@ -97,6 +97,13 @@ export enum WebcatErrorFile {
 }
 
 //
+// Internal errors — signal a WEBCAT bug. User should file an issue.
+//
+export enum WebcatErrorInternal {
+  UNEXPECTED = "ERR_WEBCAT_INTERNAL_UNEXPECTED",
+}
+
+//
 // Group wrapper (replacement for namespace WebcatErrorCode)
 //
 export const WebcatErrorCode = {
@@ -108,6 +115,7 @@ export const WebcatErrorCode = {
   Headers: WebcatErrorHeaders,
   URL: WebcatErrorURL,
   File: WebcatErrorFile,
+  Internal: WebcatErrorInternal,
 } as const;
 
 //
@@ -121,7 +129,8 @@ export type WebcatErrorCodeAny =
   | WebcatErrorCSP
   | WebcatErrorHeaders
   | WebcatErrorURL
-  | WebcatErrorFile;
+  | WebcatErrorFile
+  | WebcatErrorInternal;
 
 //
 // Error object
--- a/extension/src/webcat/response.ts
+++ b/extension/src/webcat/response.ts
@@ -254,11 +254,16 @@ export async function validateResponseContent(
     } catch {
       /* filter may already be closed */
     }
-    errorpage(
-      details.tabId,
-      fqdn,
-      new WebcatError(WebcatErrorCode.File.MISMATCH),
-    );
+    // onerror fires when the request is aborted from outside the filter (e.g.
+    // a header-validation listener already returned {cancel: true} and called
+    // errorpage with a specific code).
+    if (where !== "onerror") {
+      errorpage(
+        details.tabId,
+        fqdn,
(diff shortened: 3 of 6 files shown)


jupenur:
Nice! Looks like a solid approach, no objections from me.

lsd-cat:
Changed approach and adapted to the refactoring. The code now wraps at listener registration in `RequestHandler`: any listener that throws forces the response to cancel: true, which seems more readable and covers all the listeners in one place. There's also a small fix slipped in that prevents throws on base64 parsing in manifests.

I'm not sure why we had the `details.state` check, but I don't think it should be a case under good circumstances after the refactoring, but I might be wrong.

extension/src/browser/requests.ts
```diff
@@ -177,45 +206,97 @@ export class RequestEvent<T extends RequestDetails> extends Event {
...
 @permissions.require("webRequest")
 @permissions.require("webRequestBlocking")
-// eslint-disable-next-line @typescript-eslint/no-unsafe-declaration-merging
 export class RequestHandler extends EventTarget {
   readonly #details = new Map<string, RequestDetails>();
+  readonly #wrapped = new WeakMap<object, EventListener>();
 
   #currentListeners: RegisteredListeners = {};
 
+  /** @group Methods */
```
jupenur, line 237: Nit: the group tags are unnecessary here.

extension/src/webcat/handler.ts
```diff
@@ -226,9 +242,11 @@ export class WebcatRequestHandler extends RequestHandler {
...
     }
 
-    // Skip non-enrolled requests
+    // #onRequest attaches state to every non-extension request, so missing
+    // state means the request stage was lost: don't let the response through
     if (!details.state) {
-      return;
+      logger.addLog("error", `Request state lost for ${details.url}; cancelling`, details.tabId, getFQDN(details.url)); // prettier-ignore
+      return blockingResponse.set({ cancel: true });
     }
```
jupenur, line 232: The original check here is meant to allow ignoring non-enrolled origins, which might end up here immediately after startup when `bindAll` has been called but `bind` hasn't. It's an extremely rare thing, but I don't think we should block.

The flow is like this:

1. The WEBCAT background service worker is terminated after a period of inactivity
2. `example.com` (non-enrolled) issues a request and starts sending it
3. `lsd.cat` (enrolled) issues a request, which wakes up the WEBCAT background service worker
4. WEBCAT synchronously calls `bindAll` and starts loading enrollments from the asynchronous database
5. `example.com` receives the headers for its request; `#onHeaders` fires
6. Originally: `#onHeaders` ignores `example.com`; with the change: `#onHeaders` blocks `example.com`
7. WEBCAT finishes loading enrollments and calls `bind`

jupenur (changes requested):
Overriding the `addEventListener` implementation and wrapping handlers feels like an unnecessarily complicated approach. Could we instead flip the default value of `BlockingResponse.cancel` to `true` and force all handlers to explicitly set it to `false` to signal successful completion?

If you already looked into that approach and noticed something that makes it not work, feel free to dismiss this review.

lsd-cat:
I'd be fine in doing the flip instead. Brainstorming the approcahes, the reason for not choosing is that it works if WEBCAT is the only listener, but the API docs point embedders at `WebcatRequestHandler.addEventListener`, and all listeners on an event share one `BlockingResponse`, so as I understand it, another listener finishing after ours could override a cancel.

jupenur:
Good point. Here's another idea -- currently the API works like this:

```js
handler.addEventListener("beforeheaders", async (event) => {
  using blockingResponse = event.blockingResponse;
  /* ... */
});
```

Each event handler gets the same `BlockingResponse`. If we flip the default value of `cancel`, it becomes the event handler's responsibility to know how it should deal with possible earlier handlers' responses. What if we change it slightly, so that instead we give each handler a fresh `BlockingResponse` instance initialized to default values? Like so:

```js
handler.addEventListener("beforeheaders", async (event) => {
  using blockingResponse = event.createBlockingResponse();
  /* ... */
});
```

Each handler would then work independently, and it would become the dispatcher's responsibility to merge the responses. Currently the dispatching side looks like this:

```js
const event = new RequestEvent("beforerequest", details);
this.dispatchEvent(event);
return await event.blockingResponse.ready();
```

With the change we'd have something like this instead:

```js
const event = new RequestEvent("beforerequest", details);
this.dispatchEvent(event);
return await event.ready();
```

Here `event.ready()` would await for all created `BlockingResponse` instances, then merge the instances and return the result. And `cancel === true` would take precedence over everything else, so that if even a single listener cancels, the result would be canceled.

lsd-cat:
I like it, works for me, I'm gonna test and report back :)

extension/src/browser/requests.ts
jupenur, line 121: Now that we have `createBlockingResponse` and `BlockingResponse` is instantiated per listener, there's an opportunity to simplify this. I don't think we need the getter or the scope tracking anymore; the simplest possible `[Symbol.dispose]()` method should suffice:

```
/**
  * Implements the {@link Disposable} interface.
  */
[Symbol.dispose]() {
  this.#resolve(this);
}
```

Getting rid of the getter would also allow ditching the TypeScript downleveling workaround that's below this class declaration.

jupenur: This part we can just remove if we simplify the disposable implementation.

extension/src/browser/requests.ts
```diff
@@ -164,17 +168,47 @@ export class RequestEvent<T extends RequestDetails> extends Event {
...
     this.details = details;
   }
+
+  /**
+   * Creates a {@link BlockingResponse} for this event. Each listener creates
+   * its own before its first await expression, so that listeners respond
+   * independently of each other; see {@link ready} for how the responses are
+   * combined.
+   *
+   * @returns A new BlockingResponse that cancels the request until the
```
jupenur: Nit: s/until/unless/

jupenur (approved):
Much nicer now I think. The only comments are on some additional cleanup this change allows, but we can also do that separately if you prefer.

extension/src/browser/requests.ts
```diff
@@ -65,32 +65,39 @@ export type RequestDetails =
   | ErrorOccurredDetails
   | CompletedDetails;
 
+// Firefox ESR lacks Symbol.dispose. The `using` helper falls back to this
+// registered symbol, so define it before the class key below is evaluated.
+(Symbol as { dispose: symbol }).dispose ??= Symbol.for("Symbol.dispose");
```
jupenur: I was assuming TypeScript downleveling would handle this, but I guess not? I feel like we shouldn't pollute the global symbol object here, because in a library deployment that could interfere with other polyfills.

Can we instead define the method as `[Symbol.dispose ?? Symbol.for("Symbol.dispose")]()`?

lsd-cat: Done but via a typed const instead of inline, because it seems TypeScript won't accept using or implements Disposable with the expression as the key.

PR #179, opened 2026-05-19
Test manifest update logic

lsd-cat:
Addresses https://github.com/freedomofpress/webcat/issues/171

Adds an integration test to check websites can signal a new manifets version, and that the extension can load the new one at refresh. There's the practical problem that navigating to the same URL might load from cache and skip the header, while a refresh forces an actual reload to happen. I don't think this is specifically a WEBCAT problem, rather when we write documentations we should just clarify how to handle upgrades in the best way. For instance by either redirecting to a new path (like `?v2`) or, by regularly fetching async and force the refresh from the application itself. Maybe we'll revisit this in the future, but for now the current state is better than nothing.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -6,7 +6,7 @@ import canonicaljson
 import hashlib
 
 from helpers import Browser, UpdateServer, Server, TorBrowser, generate_ssl_cert
-from sigsum import generate_bundle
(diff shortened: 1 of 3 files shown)


jupenur (approved):
LGTM

PR #180, opened 2026-05-19
Mock CometBFT and ICS23 instead of update and db

jupenur:
This PR improves integration test coverage by using the real WebcatDatabase and update implementations and only mocking CometBFT, ICS23, and the update server instead. The goal is to eventually enable testing of scenarios like the in-memory-cache handling on update in #137, although that still requires some additional work.


lsd-cat:
Thanks, this was a big gap in the "integration" part of integration tests :)

PR #181, opened 2026-05-19
Fix benchmarks

jupenur:
The CSP header in the benchmarks was out of sync and was causing integrity errors. This PR establishes the EXPECTED_CSP constant as a shared source of truth.


lsd-cat:
Good catch, no objections!

Issue #172, opened 2026-05-06
Test delegation scenario


lsd-cat:
Similarly to https://github.com/freedomofpress/webcat/issues/171, there's harness to support WEBCAT delegation but likely not the full implementation: the idea is that a server could specify via an `x-webcat-delegation` another hostname. The extension would lookup in the list and match the enrollment information. If the policy hash is equal to the current domain, then we could surface to the user the information.

The idea is for self hosted webapps, to trust developers and non hosters directly and not just through auditing. If someone runs securedrop, and we hosted a canonical policy at canonical.securedrop.org, they could expose that, and we could tell the user that this istance is running a securedrop using our policy, and thus signed by us. I think UX wise it's a difficult concept, and might not be a direct indicator, but in an informational menu or popup could be useful for advanced users.

lsd-cat:
Completed in #182. It seems to actually be working, though the UI is not very useful.

PR #182, opened 2026-05-20
Cleanup delegation and add integration test

lsd-cat:
This adds a test according to https://github.com/freedomofpress/webcat/issues/172. The logic seemed fine already, with just a little bit of cleanup. Since for now the only effect is in the UI tooltip, the workaround for testing is to add a log line int he extension when that happens and monitor for events there. As a side consequence, this PRs bring the ability to monitor the extension own console, which I think could be useful for other tests as well.


jupenur (approved):
Looking good. The extension console access is definitely going to come in handy.

PR #184, opened 2026-05-26
Fix cache handling

jupenur:
Browser caches were not being cleared properly. With this PR, all caches are cleared every time 1) the extension starts 2) a manifest is updated 3) an integration error occurs 4) or a new site is enrolled. The performance hit is non-trivial but [there doesn't appear to be a way to make the `browsingData.remove` API work otherwise](https://github.com/freedomofpress/webcat/issues/137#issuecomment-4541426567).

Tests, for the most part, no longer use `sleep`. Instead there are methods to wait for server events. The test update server exposes an API to reschedule an update. A new test uses that API to check for correct cache behavior on update. The test app uses fixed ports 8080 and 8443 instead of random ports.

### DIFF
--- a/extension/src/webcat/listeners.ts
+++ b/extension/src/webcat/listeners.ts
@@ -437,7 +437,7 @@ export async function installEnrolledListeners(
     ids: registeredFqdns.filter((fqdn) => !fqdns.includes(fqdn)),
   });
 
-  clearBrowserCaches(newFqdns);
+  await clearBrowserCaches(newFqdns);
 
   console.log(
     `[webcat] installEnrolledListeners: registered listeners for ${fqdns.length} FQDN(s)`,
--- a/extension/src/webcat/listeners.ts
+++ b/extension/src/webcat/listeners.ts
@@ -19,7 +19,12 @@ import {
 } from "./response";
 import { errorpage } from "./ui";
 import { retryUpdateIfFailed } from "./update";
-import { getFQDN, isExtensionRequest, isNewerSemver } from "./utils";
+import {
+  clearBrowserCaches,
+  getFQDN,
+  isExtensionRequest,
+  isNewerSemver,
+} from "./utils";
 
 function commitVerifiedOrigin(fqdn: string, holder: OriginStateHolder): void {
   if (holder.stale) {
@@ -214,7 +219,7 @@ export async function beforeHeadersListener(
         case "sharedworker":
         case "audioworklet":
         case "paintworklet":
-          await hookResponseContent(details);
+          hookResponseContent(details);
       }
       break;
     }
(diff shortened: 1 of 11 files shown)


lsd-cat:
> Browser caches were not being cleared properly. With this PR, all caches are cleared every time 1) the extension starts 2) a manifest is updated 3) an integration error occurs 4) or a new site is enrolled. The performance hit is non-trivial but [there doesn't appear to be a way to make the `browsingData.remove` API work otherwise](https://github.com/freedomofpress/webcat/issues/137#issuecomment-4541426567).
> 

Thanks for all the work and debugging here! As in the issue, in the long term it would be nice to have this resolved by Firefox. In the short term, I was thinking if we could workaround it somehow by using cache controls headers. We could inject `Clear-Site-Data` in the main frame response, and also add  `Cache-Control: no-cache` in the request. I think it could work, with the exception of when a main_frame is entirely served from cache and not even revalidated, which would mean that none of the listeners fire. We can force a refresh that bypasses the cache in this cases, but I feel like complexity might be increasing a lot.

Also, I can't find a useful attack for it, but clearing the full cache has its own side channel since it effects also other domains.

jupenur:
`Clear-Site-Data` was my first thought for a workaround, too, but it can only solve a subset of cases where we need to clear caches. For example, when an integrity error occurs, we could inject `Clear-Site-Data` in the `DENIED` response, but if the requesting origin is a third party, nothing gets cleared. The header only works when the requesting origin is same-origin with the target of the request. Waiting for the next navigation is not a great option because it leaves a window of attack that an actively malicious server can easily exploit through a third-party origin.

jupenur:
Tests should be passing now and this is ready for review. I'll see if I can make the `Clear-Site-Data` trick work somehow, but I'll do that in a separate branch. This PR just clears all caches every time, but the cache-clearing logic is now all in one place so it's easier to change.

Edit: Tried out some ideas, could not make the `Clear-Site-Data` trick work.

test/tests.py
```diff
@@ -105,95 +115,99 @@ def setdiff(a: list, b: list):
...
+        ], NON_FRAME_PATHS,
         id="corrupted_wasm_frame_test"),
 
 ], indirect=["root"])
-def test_webcat(browser, in_frame, server, expected, logs, errors, rejections, addon_path, dnsnames, non_enrolled_dnsnames):
+def test_webcat(browser, in_frame, server: Server, update_server: UpdateServer, expected, logs, errors, rejections, paths_to_wait, addon_path, dnsnames, non_enrolled_dnsnames):
     logs, errors, rejections = logs.copy(), errors.copy(), rejections.copy()
     browser.install_extension(addon_path)
-    sleep(7)
+    update_server.wait_for_update()
```
lsd-cat, line 202: Does this ever race? `install_extension` would trigger an update to run immediately, and that could happen before `wait_for_update` is ready, and so the test might up ending for the next scheduled update. However, I did not manage to naturally observe the case to happen locally, but I wonder if it might in CI

extension/src/webcat/listeners.ts
```diff
@@ -408,34 +413,32 @@ export async function installEnrolledListeners(
...
+      };
+    }),
   );
   // Remove the content scripts whose fqdn is no longer enrolled
   await browser.scripting.unregisterContentScripts({
-    ids: scriptIds.filter((id) => !fqdns.includes(id)),
+    ids: registeredFqdns.filter((fqdn) => !fqdns.includes(fqdn)),
   });
 
+  clearBrowserCaches(newFqdns);
```
lsd-cat: Should we await here too? I think it's unlikely to lose the race here, but since we do for the other calls maybe it make sense to enforce the ordering?

lsd-cat (approved):

lsd-cat:
Thanks so much for this! There's a couple questions but feel free to dismiss if you don't think they actually apply. The test improvement logic is really cool! I otherwise hope we get a cache API fix before going stable but as you found out there's no choice in the meantime. Thanks for also testing and ruling out the `Clear-Site-Data` experiments

test/tests.py
jupenur: There's an opportunity for a race there for sure, but in practice it seems the extension starts up much slower than the test progresses to this wait. I think the original 7 second sleep already hints at that pretty strongly. If we ever hit this, we can fix it then. The symptom would be the test hanging for an hour, so it should be pretty obvious.

extension/src/webcat/listeners.ts
jupenur: Good catch. I agree it probably doesn't make any difference in practice, but let's await anyway. At least it's more consistent that way.

PR #187, opened 2026-06-02
Major dependencies bump

lsd-cat:
Mostly to silence the alerts, and since there were some braking changes, migrate now so new code conforms to that. Bumped TypeScript 5 -> 6.0.3, ESLint 9 -> 10.4.1, @types/firefox-webext-browser 120 -> 143.0.0, and  vitest 3 -> 4.1.8.

Let's see if integration tests pass on CI.


lsd-cat:
`vite-plugin-web-extension` was keeping a bunch of things behind, but it seems like we didn't use it so I just dropped it. Everything else is just adjustments to comply with the new versions.

jupenur (approved):

Issue #186, opened 2026-05-28
Improve error pages


lsd-cat:
Per TBB request, it would be nice if we displayed any WEBCAT related error under the URL that triggered it, or if that's not possible, that we at least report that in the error body.

In theory, we have two options to replace the page under the original domain: via interception, or overwriting the DOM. However, there's a few limitations: via interception, we can directly replace the main_frame only during the main_frame response. If the failure is in a subresource, or a worker, replacing the subresource with an error wouldn't be effective. We'd have then to trigger a refresh and mark somewhere that that refresh should be overwritten with an error page.

Overwriting the DOM can be effective in a graphical sense. My original goal for the redirect instead of the rewrite is that navigating away from the origin will stop the execution of JS too, while a graphical change wouldn't. For instance, we could overwrite the DOM, but then a page might have left a `setTimeout` operation or similar that overwrites our page. It's not per-se a security that, as that JS would be verified and transparency logged, but it isn't great in general that something can mess with extension or browser generated error pages.

Overall, I don't think either solution is better than now, but I'll look how Firefox implement the errors or if there's anything we can do on the extension side re: scheduled execution and similar. Depending what Tor wants, we either can get away by adding the URL in the body, or we need to ask for some additional browser integration.

PR #188, opened 2026-06-03
Improve error pages

lsd-cat:
Following https://github.com/freedomofpress/webcat/issues/186. We cannot really override a page securely as Firefox does, as they use internal functions that are not exposed to web extensions. The safe action is to navigate away, and the extension page is what we have. I have thus spent some time to better integrate it among the other Firefox errors (the WEBCAT icon is to signal that this is an extension error, cause Firefox doesn't like it when people get blocking errors and they could be blamed).

The page now surfaces the host in the main body, and the error code under "Advanced". If the error is related to a file that is broken or missing from manifest, it will surface the affected path (which could help development).

As @jupenur suggested, we still need a better debugging interface for administrators trying to enroll and having integrity issues, but this is a start.

image
image

### DIFF
--- a/extension/pages/error.css
+++ b/extension/pages/error.css
@@ -1,0 +1,9 @@
+body.webcaterror .title {
+  background-image: url("../icons/light/webcat-error.SVG");
+}
+
+@media (prefers-color-scheme: dark) {
+  body.webcaterror .title {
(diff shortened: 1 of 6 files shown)


jupenur (approved):
Awesome! Looks great!

PR #190, opened 2026-06-03
Add auto rerun to integration tests

lsd-cat:
Add `pytest-rerunfailures` to automatically rerun tests in case of flaky CI


lsd-cat:
The fact that this passed on the first run seems like a good omen

jupenur (approved):
Good stuff. Thanks!

PR #191, opened 2026-06-04
Cache partitioning

jupenur:
There's a lot of stuff crammed into this PR, so bear with me. The basics:

1. The `origins` cache as well as `nonOrigins` and `tabs` globals are now partitioned by first party origin and incognito status. That's done by using composite cache keys implemented using the new `CacheKey` function. This part is largely trivial.
2. There is a new utility function `getFirstParty` to help determine the correct cache partition for a webRequest details object. This is where things get a bit hairy, but there are comments in the function to help navigate it.
3. There are three cases where the first party origin cannot be determined based on a webRequest details object alone:
    1. When the request is related to a `SharedWorker`; in this case we patch the `SharedWorker` constructor to include the first party origin in the worker script's URL fragment.
    2. When the request is related to a `ServiceWorker`; piggy-backing on URL fragments doesn't work because the [URL fragment is explicitly cleared by the browser](https://w3c.github.io/ServiceWorker/#start-register-algorithm); instead we completely block service worker registration in frames and assume the first party origin for any service worker is the origin of the service worker.
    3. When the request is related to a `Worker`; this case should be easy, but due to [BMO#2048884](https://bugzilla.mozilla.org/show_bug.cgi?id=2048884) we need to do similar patching as with shared workers.
4. First party origins are cached per-request just like pending origins; that way `getFirstParty` doesn't have to make a trip to the async `webNavigation` API repeatedly.

Some hairy details:

1. To piggyback on the URL fragment of a worker script, we use AES-GCM: the encryption key is only available in the background page and the fragment leaking to web content reveals nothing that could be privacy sensitive. GCM is an authenticated operating mode, so web content cannot spoof a first party origin. A fragment can still be copied from one URL to another and reused, but I was not able to come up with a scenario where that would be a security issue.
2. There's still the risk that I've missed some case where the first party origin cannot be determined. I've tried to make `getFirstParty` fail securely by emitting a unique origin in that case, effectively disabling the cache. If that happens, the function also logs a warning.

Some additional stuff in this PR:

1. With c897807, the CI pipeline updates NoScript automatically in case there is a new version of it not bundled with TBB yet. I added that because of the `AudioWorklet` bug in NoScript.
2. I overhauled the JavaScript hooking mechanism pretty thoroughly. There's now an `updatableHook` function that handles delivering async data to a hook in a reusable manner.
3. I started using JSDoc comments in the hook library because they're surfaced nicely in IDEs. I think we should start using basic JSDoc syntax everywhere.
4. There's also some refactoring going on the listeners: `pendingOrigins` and the first party origin cache was merged to a single map and the listeners updated to match that. While I was making changes to the listeners, I eliminated some unnecessary code (fb9fc8d).


extension/src/webcat/hooks/workers.ts
```diff
@@ -0,0 +1,252 @@
...
+  unwrap,
+  updatableHook,
+} from "./core";
+import { connectEventListeners, hookEventProperty } from "./events";
+
+/**
+ * Hooks the SharedWorker API to expose information about the first-party
+ * origin to the webRequest API.
+ */
+export const sharedWorkerHook = updatableHook(
```
lsd-cat, line 18: I wonder why this is async? The SharedWorker constructor is synchronous, and the first party is substituted into the hook by getHooks before injection, so it's known synchronously at construction. The whole message relaying increases complexity quite a bit, and could break postMessage beside just the messageerror, as it seems also a MessagePort would get dropped, and clonable stuff like ArrayBuffers gets cloned twice.

extension/src/webcat/genhooks.ts
```diff
@@ -1,20 +1,39 @@
...
+
+export async function getHooks(
+  type: hooksType,
+  wasm: string[],
+  firstParty: string,
+  sameOrigin: boolean,
+) {
+  // This just patches the script string dynamically,
+  // adding per-origin WASM hashes and FPO hints
+  const iv = crypto.getRandomValues(new Uint8Array(new ArrayBuffer(96)));
```
lsd-cat: nit: it's irrelevant but I think should be 12 bytes (96 bits)

extension/src/webcat/utils.ts
```diff
@@ -84,3 +85,78 @@ export async function clearBrowserCaches(fqdns: string[]) {
...
+      const markerIndex = url.lastIndexOf("#");
+      if (markerIndex !== -1) {
+        try {
+          const efpo = Uint8Array.fromBase64(url.substring(markerIndex + 1), {
+            alphabet: "base64url",
+          });
+          const fpo = await crypto.subtle.decrypt(
+            {
+              name: "AES-GCM",
+              iv: efpo.slice(0, 96),
```
lsd-cat: nit: same as above, I think this can be 12 bytes

extension/src/webcat/genhooks.ts
lsd-cat: It is also my understanding that the worker themselves will be cached based on their url, including the fragment. In a quick local test, it seems that two instantiation of the same sharedworker from the same first party origin will lead to two workers instead of a worker with two connections. I wonder if there's specific benefits in randomizing per connection, or we could avoid the performance penalty/duplication of this by reusing the same iv per partition key (and thus the same plaintext/ciphertext, as we can't safely just reuse ivs). We could for instance deterministically generate the iv with a key derivation function using as input firstparty+secretkey.

lsd-cat (commented):
Thank you! This is a massive and impressive amount of work and was the biggest blocker on our side for integration! It's unavoidably complex, but looks great. Hope my comments make sense, I'm gonna take a few more days of testing and understanding but I really like the shape of it.

extension/src/webcat/hooks/workers.ts
jupenur: `getHooks` injects synchronously for `hooksType.page` but not for `hooksType.content_script`. For the latter, the script is injected twice to ensure timely injection: first without the first party or WASM hashes using `registerContentScripts` [here](https://github.com/freedomofpress/webcat/pull/191/changes#diff-9eaf967174c149d0bb2bc62e874cd5bb3fb7d1270b3d248d6a1db589bfdc6fd9R375-R386), where it's guaranteed to run at `document_start`, then via `executeScript` [here](https://github.com/freedomofpress/webcat/pull/191/changes#diff-9eaf967174c149d0bb2bc62e874cd5bb3fb7d1270b3d248d6a1db589bfdc6fd9R136-R145), where the first party and WASM hashes are available. The `executeScript` API does not guarantee that the script runs before document load even when called from a blocking `webRequest` listener.

extension/src/webcat/genhooks.ts
jupenur: Whoops, yeah.

jupenur: Yeah this is a good catch. SharedWorkers definitely need a deterministically generated URL fragment, not just for performance and caching, but for correct functionality. I had this right before switching to the AES-GCM approach, but then apparently forgot about it.

I'll switch to deterministic IVs. HKDF with a global random salt should be good enough to prevent maliciously generated collisions.

extension/src/webcat/hooks/workers.ts
lsd-cat: Oh I see why it is async, thanks for the explanation! Though I think the current implementation would still drop the "transfer" optional parameter of [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) silently?

jupenur: Oh yeah that's right. I was trying to avoid having to separately hook the MessagePort prototype, but that might have been a bad call. I'll think it over and maybe switch the approach.

extension/src/globals.ts
```diff
@@ -1,17 +1,49 @@
...
+import { CacheKey, LRUCache, LRUSet } from "./webcat/cache";
 import { WebcatDatabase } from "./webcat/db";
 import { stringToUint8Array, Uint8ArrayToBase64Url } from "./webcat/encoding";
 import { OriginStateHolder } from "./webcat/interfaces/originstate";
 
-export const origins = new LRUCache<string, OriginStateHolder>(lru_cache_size);
-export const pendingOrigins: Map<string, OriginStateHolder> = new Map();
-export const nonOrigins = new LRUSet<string>(lru_set_size);
-export const tabs: Map<number, string> = new Map();
+export type CachePartition = { firstParty: string; incognito: boolean };
```
lsd-cat, line 7: I was wondering, do we have a way to purge the incognito entries only when all the incognito tabs get closed? Because I think this is currently inconsistent with the incognito behavior where the context gets destroyed and re-created on each sessions.

There could also be a side-channel were, since the cache total is shared, one could fill the cache and then correlate via the evictions between incognito and non-incognito? It feels a bit elaborate, but maybe the direction could be to have two separate caches instead of a flag?

extension/src/webcat/response.ts
```diff
@@ -144,11 +146,14 @@ export async function validateResponseHeaders(
       details.tabId,
       fqdn,
     );
-    origins.delete(fqdn);
+    const cachePartition = {
+      firstParty: await getFirstParty(details),
```
lsd-cat: Could we pass this down when calling `validateResponseHeaders` instead of recomputing? Mostly because the call also depends on browser state (`getAllFrames`) and I worry there could be race conditions where it gives different results depending on the stage.

extension/src/globals.ts
jupenur: > do we have a way to purge the incognito entries only when all the incognito tabs get closed

No 100% reliable way that I'm aware of, but we could listen for `browser.windows.onRemoved` query for remaining windows, and purge if the number of incognito windows is 0. The part where that fails is if you open a new incognito window before the query runs: the browser would end the incognito session and start a fresh one, but WEBCAT would only see the newly opened incognito window and continue the original session. That's not a huge problem though. I'll look into implementing it.

> since the cache total is shared, one could fill the cache and then correlate via the evictions between incognito and non-incognito

What would an attack look like there? An attacker could force-evict incognito caches from a non-incognito context and vice-versa, but is there something they can learn from that? What's there to measure? And if an attacker fills up the cache and somehow detects evictions triggered by the user navigating to legitimate sites, they still wouldn't be able to detect whether the navigation is in incognito or non-incognito.

extension/src/webcat/response.ts
jupenur: We could and should. Will fix.

extension/src/globals.ts
lsd-cat: > What would an attack look like there? An attacker could force-evict incognito caches from a non-incognito context and vice-versa, but is there something they can learn from that? What's there to measure? And if an attacker fills up the cache and somehow detects evictions triggered by the user navigating to legitimate sites, they still wouldn't be able to detect whether the navigation is in incognito or non-incognito.

Mostly I was thinking of same-site correlation between incognito and non. The same origin could fill the cache in non incognito, then evict a known number in incognito, and then count the evictions from the original non incognito tab.

jupenur: Do you mean the attacker could use it to detect if their site is loaded in incognito and non-incognito at the same time? Wouldn't there be plenty of simpler ways to do that even without WEBCAT?

lsd-cat: Yeah you are right that's outside the threat model of it, I was overthinking it :)

extension/src/webcat/hooks/workers.ts
jupenur: 6486cf1 hooks the MessagePort prototype and should fix the postMessage handling.

extension/src/webcat/response.ts
jupenur: Done in 38aa1d1

extension/src/globals.ts
jupenur: > do we have a way to purge the incognito entries only when all the incognito tabs get closed

Done in 3cee51f

extension/src/webcat/response.ts
```diff
@@ -232,37 +233,38 @@ export async function validateResponseContent(
...
     // build a blob later. If the data is the hook marker, replace it with the WASM hooks, and if it is the
     // end marker, flush all buffered data
     if (arraysEqual(hookMarker, new Uint8Array(event.data))) {
-      const hooks = getHooks(hooksType.page, manifest.wasm);
-      source.push(stringToUint8Array(hooks).buffer);
+      const hooks = getHooks(
+        hooksType.page,
+        manifest.wasm,
+        cachePartition.firstParty,
+        cachePartition.firstParty === details.originUrl,
```
lsd-cat: I think there's an inconsistency with the other similar check `          cachePartition.firstParty === new URL(details.url).origin`, in one case the origin is the classic scheme+host, in the other is the full URL.

extension/src/webcat/hooks/workers.ts
```diff
@@ -0,0 +1,326 @@
...
+          originalPostMessage.call(unwrap(this)[internal].instance, ...args);
+        } else {
+          let options = args[1];
+          if (options && global.Symbol.iterator in options) {
+            const transfer = options as Transferable[];
+            options = new global.Object();
+            options.transfer = transfer;
+          }
+          args[0] = global.structuredClone(args[0], options);
+          unwrap(this)[internal].messages.push(args);
```
lsd-cat, line 59: If I understand correctly, this is queuing messages when the instance is not ready yet, same as it is done in the workerHook, however there's no equivalent here where the messages then gets replayed when ready.

extension/src/webcat/response.ts
jupenur: Fixed in 00f33ad. On a related note, the number of overloads we have for the term "origin" is slightly problematic. I'm regularly getting confused over which thing the word "origin" is actually referring to 🙄

extension/src/webcat/hooks/workers.ts
jupenur: Fixed in d11d750. Well spotted.

lsd-cat (approved):
Thanks so much for all the work and the back and forth with the changes! I'll wait for CI to pass and then merge :)

PR #194, opened 2026-06-17
Preserve history on integrity error

### LINKED ISSUE #193: Error page can replace the wrong page in browser history
When an integrity error is triggered for a `main_frame` response before the browser commits to navigation, the error page replaces the wrong page. When that happens, clicking "Go back" on the error page seemingly takes you back *two pages*, because the error page replaced the page the tab was navigated to *before* the error instead of the page that was being navigated to.

jupenur:
Fixes #193

### DIFF
--- a/extension/src/webcat/listeners.ts
+++ b/extension/src/webcat/listeners.ts
@@ -104,6 +104,8 @@ export async function headersListener(
     return {};
   }
 
+  const isFrame = FRAME_TYPES.includes(details.type);
+
   // pendingOrigins is populated by requestListener
   let originStateHolder = pendingOrigins.get(details.requestId);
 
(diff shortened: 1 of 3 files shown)


lsd-cat (approved):
lgtm! Thanks, tested on everything at https://integrity-error.demo.webcat.tech/ and now back navigation works on all type of errors there.

jupenur:
There's something wonky with the integration tests. I was expecting `tbb_safer` to fail, but Firefox latest should pass. Instead it silently skipped a bunch of tests after the 76% mark, jumping to 85% with no output, and then started failing. I'll dig into that tomorrow.

image

jupenur:
Firefox added a new ToS dialog that was messing with tab focus and breaking tests and I assume the missing output was a GitHub glitch. Pushed a fix and will merge once tests pass. Still expecting `tbb_safer` to fail, but that's unrelated.

Issue #185, opened 2026-05-27
i18n/i10n


redshiftzero:
We should i18n the app and load the user visible strings from a file

Eventually, these are the supported languages for Tor Browser: https://support.torproject.org/tor-browser/getting-started/languages/

PR #196, opened 2026-06-18
Implement i18n and localize to Finnish

jupenur:
Internationalize using the browser.i18n API and add Finnish language strings. This PR covers the i18n part of #185.


jupenur:
Test failures are all unrelated. See https://github.com/freedomofpress/webcat/pull/194#issuecomment-4741652649.

lsd-cat (approved):
Thanks a lot! Looks great and will add italian in solidarity :)

PR #207, opened 2026-07-09
Ensure the end marker never leaks

### LINKED ISSUE #206: endMarker sometimes leaks into response bodies
The `endMarker` token sometimes, very rarely, leaks into response bodies. It can occasionally be seen in integration tests (when run locally in non-headless mode), but when it happens to the HTML document, it does not cause a test failure, because the rest of the document is still there and parses correctly.

This needs to be investigated and fixed. Tests should also be made to fail without retries when the marker leaks, because knowledge of the marker may allow a server to bypass validation.

### LINKED ISSUE #208: Partially loaded content causes integrity errors
When on a slow connection, e.g. 4G or Tor, it's normal for large resources not to finish loading before you navigate to another page. When that happens, the connection is interrupted and the partial content triggers an integrity error in WEBCAT. The problem is easy to reproduce: enable throttling in Firefox's Network Monitor, then navigate to a page with multiple large images, and navigate away before all images finish loading.

#207 already fixes one corner case where a request is issued but interrupted before receiving headers. For a more complete fix, WEBCAT should also attempt to detect partial content based on the `Content-Length` header and block without an integrity error.

jupenur:
Fixes #206. I was not able to reproduce the leak I saw in #206 and so did not find the root cause. However, this PR fixes another marker-related bug, where, if the browser canceled a script request before receiving headers (for example due to navigating away from the page that issued it), the end marker would not be produced, and WEBCAT would compute the hash for the hooks instead of the empty content. The same fix also guards against the end marker leaking: if the end marker is not seen when expected, the stream filter writes an empty body. That should fully mitigate the security impact of #206.

As for tests, `test_webcat` now fails without retries if a marker is ever seen in the text content.

This also fixes #208, another case related to canceled requests: when partial response content has already been received, and the request is canceled, the response is now ignored.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -102,5 +102,5 @@ jobs:
       - name: Run integration tests
         run: |
           make test TESTARGS="--addon ../dist/webcat-extension-test.zip \
-          --headless --reruns 2 --reruns-delay 5 -k '${{ matrix.test-filter }}'"
+          --headless --reruns 2 --reruns-delay 5 --rerun-except Fatal -k '${{ matrix.test-filter }}'"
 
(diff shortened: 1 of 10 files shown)


jupenur:
Rebased and force-pushed to get a clean diff.

lsd-cat (approved):
lgtm! Thanks

PR #209, opened 2026-07-15
Speed up and harden the integration test harness

lsd-cat:
It seems like the speedup is significant, from the 2+ hours to around ~10 minutes max.

- Fix a race in `UpdateServer.wait_for_update()`: if the extension fetched `list.json` before the wait started, the test blocked until the next update alarm, 5 minutes later. 
- Cut per-test setup from ~30s to ~10s: create the Firefox profile once and clone it per test, sign the bundle once per session instead of per test, and replace geckordp's CPU-settle startup heuristic with polling the debugger.
- `destroy()` waits for Firefox to exit and release the debugger port; a failed start() kills the half-started browser.
- test_multiple_tabs / test_cache_eviction: assert the error page in any tab (Browser.find_tab()) instead of the selected one, which differs between Firefox and Tor Browser.

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -301,8 +301,6 @@ def test_multiple_tabs(browser: Browser, server: Server, update_server: UpdateSe
             f"window.open('{server.url()}');"
             f"setTimeout(() => location.href = '{server.url()}/x', 1000)"
         )
-    # Which tab ends up selected differs between Firefox and Tor Browser;
-    # the corrupted script must be blocked in whichever tab loaded it.
     assert browser.find_tab(expected), \
         f"no tab showing {expected}: {[t.get('url') for t in browser.root.list_tabs()]}"
 
@@ -353,8 +351,6 @@ def test_cache_eviction(browser: Browser, server: Server, update_server: UpdateS
             f"    location.href = '{server.url(dnsnames[1])}/console_log.png';"
             "}, 1000)"
         )
-    # Which tab ends up selected differs between Firefox and Tor Browser;
-    # the evicted-and-corrupted file must be blocked in whichever tab loaded it.
     assert browser.find_tab(expected), \
         f"no tab showing {expected}: {[t.get('url') for t in browser.root.list_tabs()]}"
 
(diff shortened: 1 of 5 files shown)


test/tests.py
```diff
@@ -301,8 +301,8 @@ def test_multiple_tabs(browser: Browser, server: Server, update_server: UpdateSe
             f"window.open('{server.url()}');"
             f"setTimeout(() => location.href = '{server.url()}/x', 1000)"
         )
-    res = browser.execute("document.body.textContent")
-    assert expected in res
+    assert browser.find_tab(expected), \
+        f"no tab showing {expected}: {[t.get('url') for t in browser.root.list_tabs()]}"
```
jupenur: Is there a good reason for changing this? Using `find_tab` slightly relaxes the condition by which the test passes, so a bug could sneak through a bit more easily. Same goes for the other test.

jupenur (commented):
Impressive! Just one question before I approve.

test/tests.py
lsd-cat: Good point, it's a shortcut to address some flakiness that I experienced (sometimes the correct tab would be already the one in focus, sometimes not). I'm gonna try to understand the root cause so it's more deterministic rather than a blind search

jupenur: Tab order should be deterministic, but if the browser errors and shows a modal, that can prevent the tab focus from switching correctly. It TBB that can happen for example if Tor fails to start.

lsd-cat: Got it thanks, I'm gonna do some more attempts to improve the logic there and ping you again when it's ready

jupenur: Looks like that's exactly what's happening. Something in this PR is messing up Tor, which triggers a modal in the browser. The Tor error also persists across profiles after it first occurs. Rebooting clears it up.

image

jupenur: `killall firefox.real` also clears up the error. I'll leave the rest of the spelunking to you :)

lsd-cat: `kill_tree` now handle better killing of tor browser, including tor and not only the browser. I've restored the original matching logic and have experienced no flakiness, so I think that was the issue.

jupenur (approved):
Awesome!

PR #210, opened 2026-07-16
Refactor, break circular dependencies, get rid of globals

jupenur:
This is a fairly big refactor that improves maintainability by getting rid of `globals.ts` and *all* circular dependencies. Circular dependencies are now disallowed via `npm run lint:check`. Package organization is slightly improved and a lot of the stuff that used to be in functions is moved to cohesive classes. There are some more JSDoc comments, too.

The biggest architectural change is the introduction of the `RequestHandler` and `ContentScript` classes that abstract away some of the hairiness of the browser APIs.

098806c contains a bugfix that should be cherry-picked if we end up not merging the whole PR: in hooked `EventTarget` methods, `this` can be `undefined` in strict mode, which was causing errors.


jupenur:
There's plenty that could still be improved, but I have to stop somewhere. I might still add more unit tests, but those shouldn't affect what's already there, so this is ready for review.

extension/src/browser/requests.ts
```diff
@@ -0,0 +1,329 @@
...
+      details: browser.webRequest._OnErrorOccurredDetails,
+    ) => this.#errorOccurred(details);
+    const completed = (details: browser.webRequest._OnCompletedDetails) =>
+      this.#completed(details);
+
+    // Add new listeners first, then remove the old ones
+    browser.webRequest.onBeforeRequest.addListener(before, { urls }, [
+      "blocking",
+    ]);
+    browser.webRequest.onBeforeSendHeaders.addListener(
```
lsd-cat, line 230: Before here we were listening only from "type": "scripts", it's my understanding that the overall logic didn't change but has mostly been re-organized, so I wonder why this change since it now intercepts a lot more?

extension/src/webcat/updater.ts
```diff
@@ -0,0 +1,288 @@
...
+      }
+      console.log(`[webcat] List updated successfully`);
+      this.dispatchEvent(new UpdateEvent(true, local));
+
+      // Success - clear failure flag
+      this.#lastUpdateFailed = false;
+    } catch (error) {
+      console.error("[webcat] Update failed:", error);
+      this.#lastUpdateFailed = true;
+      this.dispatchEvent(new UpdateEvent(false, local));
```
lsd-cat, line 208: If I read everything correctly, the event is now dispatched in all cases: success, failure and nothing to apply. I think the cache clearing function has gating afterwards, but the others do not, and so the listeners are deleted and re-added in all cases?

extension/src/webcat/updater.ts
```diff
@@ -0,0 +1,288 @@
...
+        browser.alarms.create(this.#alarm, {
+          periodInMinutes: this.#checkInterval / 60000,
+        });
+      }
+      this.dispatchEvent(new Event("scheduled"));
+    });
+
+    console.log("[webcat] Importing bundled list");
+    this.update(true)
+      .then(async () => {
```
lsd-cat: I think here it would be better if the network update run even if the local update failed?

lsd-cat:
Thanks, this is a lot of improvement! I did a general pass trying to see if there's difference in behavior from before, and then a more detailed review of the updating logic

extension/src/browser/requests.ts
jupenur: Yeah, I originally thought I'd end up moving the validation logic around some more and would need the more widely scoped listener. Turns out I didn't, but forgot to undo the scoping change.

jupenur: Fixed in 188aa2c.

extension/src/webcat/updater.ts
jupenur: Added proper gating in 593cb8f.

extension/src/webcat/updater.ts
jupenur: Agreed. Fixed in 7aae0ac.

lsd-cat (approved):
Thanks, way more readable now!

PR #213, opened 2026-07-28
Persist state in browser.storage

### LINKED ISSUE #159: MV3: move in-memory global state to browser.storage.session
If we are to migrate to manifest v3, all global state stored in memory needs to be moved to the [`browser.storage.session`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/session) storage area. Data in the session storage area is guaranteed to persist across ServiceWorker restarts and never be written to disk, meeting our needs nicely. There is a 10 megabyte quota that cannot be expanded, but it's probably enough if we're smart about how we use it.

jupenur:
Introduce new general-purpose KVStore and NamespacedKVStore classes that the extension uses to persist all non-ephemeral state. Fixes #159.

### DIFF
--- a/extension/src/background.ts
+++ b/extension/src/background.ts
@@ -67,7 +67,7 @@ requestHandler.addEventListener(
   },
 );
 
-const firstUpdate = true;
+let firstUpdate = true;
 updater.addEventListener("updated", async (event) => {
   if (!event.success && !firstUpdate) {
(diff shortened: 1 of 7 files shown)


lsd-cat (approved):
Looks goods, no comments here!

PR #214, opened 2026-07-28
Migrate to manifest v3

### LINKED ISSUE #58: Manifest v3
We should investigate if we can securely support Manifest V3 for compatibility with all browsers

jupenur:
Resolves #58.

### DIFF
--- a/extension/src/webcat/db.ts
+++ b/extension/src/webcat/db.ts
@@ -1,6 +1,6 @@
 import { NamespacedKVStore } from "../browser/kvstore";
 import { lru_cache_size, lru_set_size } from "../config";
-import { CacheKey, LRUCache, LRUSet } from "./cache";
+import { CacheKey, LRUSet, PersistentLRUCache } from "./cache";
 import { BlockMeta, Database } from "./interfaces/database";
 import { CachePartition } from "./interfaces/originstate";
 import { OriginState } from "./originstate";
@@ -9,9 +9,10 @@ import { extractHostname, extractRawHash } from "./parsers";
 const META_KEY = "block_meta";
 
 export class WebcatDatabase extends NamespacedKVStore implements Database {
-  readonly origins = new LRUCache<CacheKey<CachePartition>, OriginState>(
-    lru_cache_size,
-  );
+  readonly origins = new PersistentLRUCache<
+    CacheKey<CachePartition>,
+    OriginState
+  >(lru_cache_size, this.namespace("origins"), "session", OriginState);
   readonly nonOrigins = new LRUSet<CacheKey<CachePartition>>(lru_set_size);
   readonly enrollments = this.namespace("enrollments");
 
@@ -34,8 +35,8 @@ export class WebcatDatabase extends NamespacedKVStore implements Database {
     await this.enrollments.set(batch);
     await this.set({ [META_KEY]: meta });
 
-    this.origins.clear();
-    this.nonOrigins.clear();
+    await this.origins.clear();
+    await this.nonOrigins.clear();
 
     console.log(`[webcat] Replaced list with ${leaves.length} entries`);
   }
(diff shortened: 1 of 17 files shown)


jupenur:
This is targeting the #213 branch, that's why you don't see any tests running here. Review and merge #213 first and GitHub will automatically retarget this to main, then the tests should run.

jupenur:
Rebased

lsd-cat:
I wonder what's your opinion on keeping the cache in or our the persistent storage? I don't know how often the bg page will actually be terminated, but my impression would be that the cache would lose a lot of utility this way

jupenur:
The default timeout is only 30 seconds, so I agree we should probably persist the origin cache. Looking into it now.

jupenur:
`OriginState` objects are now persisted per session, matching the pre-MV3 behavior. I decided *not* to explicitly persist `nonOrigins` because the in-memory set is already populated via a simple storage lookup here: https://github.com/freedomofpress/webcat/blob/e6dd2617a888931c6a1acfb7b8e268715c369111/extension/src/webcat/db.ts#L72-L79

jupenur:
There was some refactoring involved in making the `OriginState` classes compatible with session storage. Namely, `OriginStateHolder` is now *gone*, and there's only one mutable `OriginState` class instead; the `LRUCache` API is now asynchronous (and for consistency so is `LRUSet`) and subclassed into `PersistentLRUCache`; and the `OriginState` class implements a `toPOJO` method that returns a plain JavaScript object that's compatible with the storage API.

lsd-cat:
This is great, thank you! I've spent some time trying to figure out the implications of using the storage, especially for risks of race conditions (not necessarily security-sensitive, but still). I'm gonna take some more to brainstorm and test more to see if I can spot and test concrete cases

jupenur:
03364e7 should resolve some of the concerns around race conditions.

extension/src/webcat/handler.ts
```diff
@@ -216,28 +251,28 @@ export class WebcatRequestHandler extends RequestHandler {
     return;
   }
 
-  #commitVerifiedOrigin(
+  async #commitVerifiedOrigin(
```
lsd-cat, line 254: I got a bit in a rabbit hole here. It is my understanding that executions are normally sequential and the extension backend is single thread, so inside the same function we should have some degree of atomicity. However, I'm unsure what the behavior is when we await and if execution is potentially passed to other contexts and then resumed when the await has completed. If that's the case, there's a a potential race here and in similar code patterns. Though it's pretty unlikely, I'd just like to double check if the reasoning holds

extension/src/webcat/db.ts
lsd-cat, line 34: This is not specific to this PR, but probably an earlier regression of when we moved out of IndexeDb. If the await/async assumption is correct, then there's a tiny window where the list is empty and an enrolled website could be skipped.

lsd-cat (approved):
Thanks, it's great to see this live! I think it's well structured and will improve mantainability, as it seems already with the lib draft PR. The comments are only about worrying whether adding awaits to code we expect to be sequential actually breaks some guarantees we expected in the past. I'm gonna do some more testing a dig a bit more into real behavior to see if they are plausible, but they are not necessarily blocking. Thanks also for making the cache atomic.

extension/src/webcat/db.ts
jupenur: If you're happy with the mutex approach in 03364e7, I could easily reuse it here. The other option is to not `clear`, but to explicitly remove only those entries that are not present in `batch`, similar to how `ContentScript.bind` works.

extension/src/webcat/handler.ts
jupenur: I reviewed each async method systematically and opened #223 with some preliminary notes. There's a handful of places we might need to adjust things, but it's probably better to do that in a separate PR.

lsd-cat: Sounds good to me, thanks!

extension/src/webcat/db.ts
lsd-cat: The mutex approach works, but I don't really have a fix preference here

PR #215, opened 2026-07-29
Move remaining browser API handlers from background.ts to WebcatRequestHandler

jupenur:
A small step towards packaging the extension as a library: include all necessary handlers in WebRequestHandler so that there's a single clear public API.


jupenur:
Rebased

lsd-cat (approved):
lgtm

PR #218, opened 2026-08-04
Package as a library

jupenur:
With this PR, `npm pack` creates a package that can be used to embed WEBCAT in other extensions. All exported interfaces are available in the package indiscriminately; we may choose to omit some later as we refine the library. There are minimal examples of using the library [in an extension here](https://github.com/jupenur/webcat-lib-test) and [in webcat-cli here](https://github.com/freedomofpress/webcat-cli/pull/12).

This PR does *not* include GitHub Actions for building the library, or anything related to publishing.

47822e3 contains a fix for an issue that was causing integrity errors on https://hacked.solutions/apps/drawio/.


extension/src/webcat.ts
```diff
@@ -0,0 +1,53 @@
...
+import { clearBrowserCaches } from "./webcat/utils";
+
+function start(options?: Partial<WebcatConfig>) {
+  const cfg = Object.assign(structuredClone(config.default), options);
+  const db = new WebcatDatabase(cfg);
+  const requestHandler = new WebcatRequestHandler(db, cfg);
+  const updater = new EnrollmentUpdater(Object.assign({ database: db }, cfg));
+
+  setIconsPath(cfg.iconsPath);
+  setPagesPath(cfg.pagesPath);
```
lsd-cat: It seems the `WebcatRequestHandler` constructor calls `getErrorPageURL()` which loads the path before we set it here

extension/src/webcat/response.ts
```diff
@@ -468,6 +473,7 @@ export class ResponseValidator {
 
   markContent(details: HeadersReceivedDetails) {
     if (PASS_THROUGH_TYPES.has(details.type)) return;
+    if (details.fromCache) return;
```
lsd-cat: Skipping the marker on fromCache responses means the worker hooks queued at onstart are not flushed and pollute the body, I've pushed a generated failing test to https://github.com/freedomofpress/webcat/tree/cached-workers-test

extension/package.json
```diff
@@ -3,10 +3,28 @@
   "type": "module",
   "version": "2.7.1",
   "description": "",
-  "main": "src/background.ts",
+  "files": [
```
lsd-cat, line 6: Should we also ship `dist/hooks/content.js` or have it configurable like the other paths?

lsd-cat (commented):
Thanks, this looks great so far. There's a few nits I would not have caught without AI help. What do you think of adding another Firefox test to the matrix of a dummy consumer running then all the WEBCAT tests? That way we'd catch lib-usage speicific regressions. In case I can open a separate PR with an attempt on that on top of this.

extension/src/webcat/response.ts
jupenur: Yup. This change was based on a wrong initial analysis of a bug. The actual fix for the thing that the cache check here tried to address is now in 9071190, already in main. I've cherry-picked your test to this PR branch and reverted the change that broke it.

extension/src/webcat.ts
jupenur: Thanks. Fixed in abb9e5a.

extension/package.json
jupenur: Yes, both. Done in fcbf195.

jupenur:
> What do you think of adding another Firefox test to the matrix of a dummy consumer running then all the WEBCAT tests? That way we'd catch lib-usage speicific regressions. In case I can open a separate PR with an attempt on that on top of this.

That's definitely a good idea. Feel free to take a stab at it.

jupenur:
Tests are failing because the main branch is broken. I'll investigate that separately of this PR.

lsd-cat:
Oh sorry about the failing test, when I merged the PRs everything was passing so I thought it could be a flake. Thanks for the other fixes as well!

jupenur:
Rebased to make the tests pull in the fixed main and, while I was at it, tidied up a bit by removing the cache fix commit and its revert. Tests should pass now.

lsd-cat (approved):

PR #219, opened 2026-08-05
Prefix i18n message names

jupenur:
This PR adds a `WEBCAT_` prefix to all i18n message names. That way we avoid collisions with existing strings in extensions that consume WEBCAT as a library.


lsd-cat (approved):
lgtm

PR #226, opened 2026-08-19
Check bundle freshness on every request

### LINKED ISSUE #221: Manifest freshness is only checked on initial load
If a manifest expires while cached, the origin state is not invalidated. Responses can be verified against an expired manifest as long as the manifest remains cached, which is in practice only limited by the length of the browsing session.

Manifest freshness should be re-checked on every request.

jupenur:
Fixes #221.


lsd-cat (approved):
lgtm!

PR #227, opened 2026-08-19
Allow same-origin redirects

### LINKED ISSUE #220: Allow frame redirections to same-origin absolute URLs
Currently, HTTP redirections are only allowed to relative URLs:

https://github.com/freedomofpress/webcat/blob/dc49fbd63344ff31cff00163b0a9e98dc12d5dfa/extension/src/webcat/response.ts#L277-L287

Same-origin redirections to absolute URLs should also be allowed: for example GitHub Pages redirects to an absolute URL when navigating to a directory path without a trailing slash; `https://freedomofpress.github.io/securedrop-protocol` redirects to `https://freedomofpress.github.io/securedrop-protocol/`.

jupenur:
Fixes #220.

It's worth noting that the URL constructor used in this PR and Firefox's internal URL parsing for Location headers don't behave exactly the same way: specifically, the URL constructor strips ASCII whitespace throughout the URL, while Firefox's internal handling only trims whitespace from the start of the URL. That means a URL like `https\t://example.com/` appears absolute to the URL constructor but actually triggers a relative redirect. This PR doesn't address that issue, because the failure mode is safe.

### DIFF
--- a/extension/src/webcat/response.ts
+++ b/extension/src/webcat/response.ts
@@ -52,17 +52,10 @@ function assertHeadersAvailable<T>(
   }
 }
 
-export function isSafeRelativeLocation(value: string): boolean {
-  value = value.trim();
-
-  // No scheme, no protocol-relative, no backslashes
-  return (
-    (value.startsWith("/") ||
-      value.startsWith("../") ||
-      value.startsWith("./")) &&
-    !value.startsWith("//") &&
-    !value.includes("\\")
-  );
+export function isSameOriginURL(target: string, base: string) {
+  const baseURL = new URL(base);
+  const targetURL = new URL(target, baseURL);
+  return targetURL.origin === baseURL.origin;
 }
 
 export class ResponseValidator {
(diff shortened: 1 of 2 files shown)


extension/src/webcat/response.ts
```diff
@@ -52,17 +52,10 @@ function assertHeadersAvailable<T>(
...
-  return (
-    (value.startsWith("/") ||
-      value.startsWith("../") ||
-      value.startsWith("./")) &&
-    !value.startsWith("//") &&
-    !value.includes("\\")
-  );
+export function isSameOriginURL(target: string, base: string) {
+  const baseURL = new URL(base);
+  const targetURL = new URL(target, baseURL);
```
lsd-cat: `new URL` can throw on malformed URLs, and I don't think we catch that anywhere. Without https://github.com/freedomofpress/webcat/pull/178 I think that could result in a WEBCAT bypass?

lsd-cat (changes requested):

extension/src/webcat/response.ts
jupenur: Good catch, thanks! Fixed in 73f7d5c.

lsd-cat: Great! I wonder if an alternative to #178 could be instead to lint for uncaught throws or errors in CI instead, I'm gonna investigate a bit! Thanks for the patch here :)

lsd-cat (approved):

PR #228, opened 2026-08-19
Fix benchmarks

jupenur:
Benchmarks were broken due to changes to test setup and WASM hooking. This PR fixes them.


lsd-cat (approved):
Tested locally, added a small fix to align the `wait_for_update` with the cold tests.

PR #230, opened 2026-08-21
Add an integration test with draw.io

jupenur:
This PR adds draw.io as a submodule and uses it in an integration test. That test caught a new [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=2065230), which commit 8f9b628 works around.


lsd-cat (approved):
tested locally, swapped the asset to check to not break in dark mode

PR #238, opened 2026-08-25
Cancel cross-origin redirects of extension requests

### LINKED ISSUE #231: Bundle fetch should not follow redirects
`bundle.json` and `bundle-prev.json` should not follow redirects, mostly to avoid SSRFs. I would be in favor of following same-site redirects: in that way hosters could have multiple valid bundles at the same time, for instance for testing/staging.

lsd-cat:
Fixes https://github.com/freedomofpress/webcat/issues/231. Sadly there's no way in the fetch API to get redirects step by step. There's either "accept" and verify the final destination or "reject" and get no info about it. So the workaround is to intercept our own requests and check that every hop is same-site. It also adds an integration test for it.


jupenur (approved):
LGTM

Issue #229, opened 2026-08-20
issues with default_fallback


lsd-cat:
https://git.legoktm.com/ set default_fallback so that all the paths would successfully url rewrite to the index page. Somehow webcat still report FILE_MISSING when opening any path that is not index, so maybe there's something off with the default_index logic.

jupenur:
A URL that ends in `/` always uses `default_index` even if that hash doesn't exist. It *should* use the fallback if the index isn't there. In @legoktm's case there's a workaround: use an absolute path for `default_index`.

jupenur:
Fixed in #239.

legoktm:
Thanks, I just tested by building the extension off of `main` and it works perfectly now \o/

Issue #195, opened 2026-06-17
default_fallback should be used only on main_frame navigation


lsd-cat:
As discussed in the WAICT repo and implemented there (https://github.com/waict-wg/waict-integrity-spec/issues/39) it seems reasonable to use the `default_fallback` only on frame navigation. A user could hit a stale URL and that shouldn't trigger an error. However, if a valid verified frame loads a subresource that is expected to exist in the current manifest, it should trigger an error instead of falling back, otherwise the server could decide which subresources to serve or not.

lsd-cat:
Implemented in https://github.com/freedomofpress/webcat/pull/239

PR #239, opened 2026-08-25
Use default_fallback only on main_frame navigations

lsd-cat:
Addresses https://github.com/freedomofpress/webcat/issues/195 to also match current WAICT spec.

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -549,7 +549,6 @@ def test_default_fallback(browser: Browser, server: Server, update_server: Updat
             "s.src = '/js/not-in-manifest.js';"
             "document.head.appendChild(s);"
         )
-    sleep(2) # wait for the error page to load
     assert "ERR_WEBCAT_FILE_MISSING" in browser.execute("document.body.textContent")
 
     # A directory path whose default_index entry is missing also falls back on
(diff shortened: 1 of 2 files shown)


lsd-cat:
Going to rebase also I'm not sure an integration tests is really useful here and it might be more of a slowdown of the suite, happy to remove it in case.

jupenur:
Might be good to fix #229 while you're at it; it's the same `else` branch that's causing it.

test/tests.py
```diff
@@ -520,6 +520,36 @@ def test_bundle_redirect(browser: Browser, server: Server, update_server: Update
...
+
+    # A subresource missing from the manifest must fail, even if its content
+    # matches default_fallback
+    with server.wait_for({"/js/not-in-manifest.js"}):
+        browser.execute(
+            "var s = document.createElement('script');"
+            "s.src = '/js/not-in-manifest.js';"
+            "document.head.appendChild(s);"
+        )
+    sleep(2) # wait for the error page to load
```
jupenur: Is this sleep necessary? `wait_for` has a built-in .5 second sleep and the error page is served from the extension so it should be pretty fast.

test/tests.py
```diff
@@ -520,6 +520,46 @@ def test_bundle_redirect(browser: Browser, server: Server, update_server: Update
...
+    assert "ERR_WEBCAT_FILE_MISSING" in browser.execute("document.body.textContent")
+
+    # A directory path whose default_index entry is missing also falls back on
+    # main_frame navigations (#229). Only the frame itself is asserted: the
+    # fallback page's relative subresources legitimately fail per #195
+    with server.wait_for({"/stale-dir/"}):
+        browser.navigate(f"{server.url(dnsnames[0])}/stale-dir/")
+    sleep(2) # extension logs arrive asynchronously
```
jupenur: Same here. Does it fail without the sleep?

test/tests.py
lsd-cat: Good catch, tried without both and it passes locally, waiting for CI.

jupenur (approved):

PR #240, opened 2026-08-26
Fail closed on malformed enrollment/manifest/signatures

lsd-cat:
Attacker-controlled input reached three uncaught throws in the verification path, each of which let the request through unvalidated: `canonicalize()` could throw,  and in a few circumstances the bundle could be empty or the signatures could not be an array. 

Does not address directly https://github.com/freedomofpress/webcat/issues/176, which I'm gonna take next, but ensure there's less uncaught surface.

### DIFF
--- a/extension/src/webcat/canonicalize.ts
+++ b/extension/src/webcat/canonicalize.ts
@@ -1,5 +1,3 @@
-// From https://github.com/theupdateframework/tuf-js/blob/38d537ea883e8bb38ee6ab17b5f59ee479d0eab2/packages/canonical-json/lib/index.js
-
 const COMMA = ",";
 const COLON = ":";
 const LEFT_SQUARE_BRACKET = "[";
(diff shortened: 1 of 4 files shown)


extension/src/webcat/validators.ts
jupenur: This is the last place using `canonicalize` instead of `tryCanonicalize`. Would it make sense to have just one function and use it everywhere?

lsd-cat: Makes sense, I've modified canonicalize to not throw and potentially return null, so that we can catch that via TS enforcement

extension/src/webcat/canonicalize.ts
```diff
@@ -16,7 +16,11 @@ function canonicalizeString(string: string): string {
 // as specified at http://wiki.laptop.org/go/Canonical_JSON. It's a restricted
 // dialect of JSON in which keys are lexically sorted, floats are not allowed,
 // and only double quotes and backslashes are escaped.
-export function canonicalize(object: object): string {
+/**
+ * JCS-canonicalizes a value. Returns null if it contains a value that cannot
+ * be encoded (e.g. a non-integer number).
+ */
```
jupenur: Maybe merge these into a single documentation comment?

lsd-cat: Done in https://github.com/freedomofpress/webcat/pull/240/changes/74ee8443e05d515e440f9dc2f0adf392fbbe4a2b

extension/src/webcat/canonicalize.ts
```diff
@@ -12,11 +12,14 @@ function canonicalizeString(string: string): string {
...
 }
 
-// Recursively encodes the supplied object according to the canonical JSON form
-// as specified at http://wiki.laptop.org/go/Canonical_JSON. It's a restricted
-// dialect of JSON in which keys are lexically sorted, floats are not allowed,
-// and only double quotes and backslashes are escaped.
-export function canonicalize(object: object): string {
+/**
+ * Recursively encodes the supplied value according to the canonical JSON form
+ * as specified at http://wiki.laptop.org/go/Canonical_JSON. It's a restricted
```
jupenur: I just noticed this link is also dead. I guess https://www.rfc-editor.org/info/rfc8785/ is what we're going for?

lsd-cat: As far as I remember I took the original `canonicalize()` from [sigstore-js](https://github.com/sigstore/sigstore-js)'s, however, they reorganized the repository and use [now a different implementation](https://github.com/sigstore/sigstore-js/blob/769a53d8713248a8bf49edfc2a5d1955b0dcc24d/packages/core/src/json.ts#L19).

However, it seems that a lot of places refer that as a source:
https://grep.app/search?q=http%3A%2F%2Fwiki.laptop.org%2Fgo%2FCanonical_JSON

I'm just gonna check if there's any discrepancy between the new ones or if they are all the same "standard". We also had a mismatch with the enrollment infra in the past.

jupenur: Here's a working archive link if we need to stick to the original reference: https://web.archive.org/web/20251209150702/http://wiki.laptop.org/go/Canonical_JSON

jupenur (approved):

PR #242, opened 2026-08-27
Enforce execution order using mutexes

### LINKED ISSUE #223: Review async code safety
See https://github.com/freedomofpress/webcat/pull/214#discussion_r3784463340. Initial notes based on commit 03364e7

✅ `browser.windows.onRemoved` listener: best-effort regardless of async
✅ `beforeframeload` listener: blocks the relevant response
⚠️ `updated` listener: asynchronously clears caches; a request could be processed after bind but before clearing caches
⚠️ `KVStore` and `NamespacedKVStore` methods: not atomic
✅ `BlockingResponse.ready`: does what it's supposed to
✅ `RequestHandler` private methods: all good
✅ `ContentScript.bind`: adds new scripts, then removes old scripts; safe
✅ `Mutex.acquire`: does what it's supposed to
✅ `LRUCache` and `PersistentLRUCache` methods: atomic via `Mutex`
✅ `LRUSet` methods: only superficially async, no await
⚠️ `WebcatDatabase.updateList`: calls `enrollments.clear`, resulting in a time window with no enrollments
✅ Other `WebcatDatabase` methods: all good
⚠️ `WebcatRequestHandler.bind`: first calls `RequestHandler.bind`, then the async `ContentScript.bind`; handlers could run before scripts are bound
✅ `WebcatRequestHandler.#initializeState`: no issues
✅ `WebcatRequestHandler.#onRequest`: no issues
✅ `WebcatRequestHandler.#onHeaders`: no issues
✅ `WebcatRequestHandler.#commitVerifiedOrigin`: not atomic; handlers could commit conflicting origins, but the impact is limited to an duplicated bundle fetch
✅ `WebcatRequestHandler.#getFirstParty`: failure modes are safe
✅ `HookBuilder` methods: no issues
✅ `BundleFetcher.awaitAll`: does what it's supposed to
✅ `OriginState.#verifyDelegation`: no issues
✅ `OriginState.verifyEnrollment`: no issues
✅ `OriginState.verifyManifest`: no issues
✅ `validateOrigin`: no issues
✅ `ResponseValidator.validateHeaders`: no issues
✅ `ResponseValidator.validateContent`: never awaits, unnecessarily async
✅ `filter.onstop`: no issues
✅ `errorpage`: no issues
⚠️ `EnrollmentUpdater`: `retryIfFailed` can run multiple times before `#lastUpdateFailed` is reset, each time calling `update` and triggering unnecessary fetches
✅ `SHA256`: no issues
✅ `clearBrowserCaches`: no issues
✅ `validateCSP`: no issues
✅ `witnessTimestampsFromCosignedTreeHead`: no issues
✅ `verifySigsumManifest`: no issues
✅ `verifySigstoreManifest`: no issues

jupenur:
Fixes #223:

 - The `update` listener is no longer responsible for clearing caches; caches are cleared in `WebcatRequestHandler.bind`, where a mutex ensures execution does not interleave with `#onRequest`.
 - A mutex in `NamespacedKVStore` operations ensures they cannot interleave with operations of the same instance. No guarantees are provided across instances.
 - `WebcatDatabase.updateList` no longer calls `enrollments.clear`
 - Multiple `EnrollmentUpdater.retryIfFailed` calls only call `update` once if the update succeeds.


jupenur:
This is probably easiest to review one commit at a time. Each commit pretty much fixes one issue. I could've opened separate PRs for each one but this way there's less to merge.

lsd-cat:
Hey, sorry for the delay here! I didn't have the continuous time needed to followup and understand, but will ensure do so by this week.

extension/src/browser/sync.ts
```diff
@@ -0,0 +1,66 @@
...
+    } else if (lock && lock === this.#lock) {
+      return this.#lock;
+    } else {
+      return new Promise<Lock>((resolve) => {
+        this.#resolvers.push([resolve, lock ?? this.createLock()]);
+      });
+    }
+  }
+
+  release() {
```
lsd-cat, line 53: It took me forever to understand, but it seems like there could be a bug here. Let's assume this test (which fails now locally for me):

```
it("admits all queued waiters sharing the installed lock", async () => {
  const mutex = new Mutex();
  const shared = mutex.createLock();
  await mutex.acquire();           // held by a different lock
  const a = mutex.acquire(shared); // queued
  const b = mutex.acquire(shared); // queued with the same lock
  mutex.release();
  await a;
  // `shared` is now installed, so `b` must be admitted too 
  // Instead it stays pending
  expect(inspect(b)).not.toContain("<pending>");
});
```


It seems that `release()` wakes only the first queued waiter, but the queue can hold several waiters that acquired with the same lock

extension/src/webcat/handler.ts
```diff
@@ -63,13 +67,16 @@ export class WebcatRequestHandler extends RequestHandler {
       },
     );
     browser.windows.onRemoved.addListener(this.#onWindowClosed.bind(this));
+
+    // Block requests until first bind
+    this.#mutex.acquire(this.#bindLock);
   }
 
-  override async bind(fqdns: string[]): Promise<string[]> {
+  override async bind(fqdns: string[]) {
```
lsd-cat, line 75: Both bind() calls acquire with the same #bindLock, so don't two overlapping binds run at the same time instead of one after the other? At startup the bundled import and the network update both trigger bind(), so the second one likely starts while the first is still registering content scripts

lsd-cat (commented):
I like the shape of this, even though it's getting harder for me to catch up with all the refactoring and have enough context to fully double check everything. The comments come from auditing from many angles with an agent and triaging the few worthwhile ones.

extension/src/webcat/handler.ts
jupenur: Fixed in bd54f10

extension/src/browser/sync.ts
jupenur: Fixed in f122117

lsd-cat (approved):
Thanks for the quick fixes! lgtm

PR #248, opened 2026-09-04
Set up API docs

jupenur:
This PR adds a bunch of documentation comments and sets up TypeDoc. API docs can be generated by running `npx typedoc` in the `extensions/` directory. This PR does not include publishing for the API docs, but a manually published preview site is available at https://jupenur.github.io/webcat-api-docs/.


extension/src/webcat/updater.ts
```diff
@@ -23,20 +36,58 @@ export class UpdateEvent extends Event {
...
+   */
   database: Database;
+  /**
+   * The set of CometBFT validators to verify enrollments against.
+   */
   validatorSet: ValidatorJson;
   /** Expected CometBFT chain ID; blocks from any other chain are rejected. */
   chainId: string;
+  /**
+   * The time in seconds between enrollment update checks. Determines how often
```
lsd-cat: nit: I think this is time in milliseconds and the same applies to `updateInterval` later. Can either change the doc or the code, but I'd say we stick to everything in the same unit, so whatever we chose should also be `fetchTimeout`

extension/src/webcat/cache.ts
```diff
@@ -71,40 +122,112 @@ export class LRUCache<K, V> {
...
+  fromPOJO(pojo: object): V;
+}
+
+/**
+ * Arguments for the {@link PersistentLRUCache} constructor:
+ *  1. limit: the maximum number of entries held by the cache
+ *  2. store: a {@link KVStore} instance to use for persistence
+ *  3. area: the name of the {@link browser.storage.StorageArea | storage area}
+ *    to use
+ *  4. (optional) depojoifier: a {@link Depojoifier} for converting presisted
```
lsd-cat: nit: typo in presisted

lsd-cat (approved):
I like the API design, the comments are just nitpicks, good to merge for me!

PR #249, opened 2026-09-04
Remove the @vitest/browser dev dependency

jupenur:
We're not using `@vitest/browser` and it's responsible for the one remaining dependabot nag, so this PR removes it.


jupenur:
Looks like a Firefox update broke something unrelated to this PR.

Edit: the fix is in #250.

lsd-cat (approved):

Issue #247, opened 2026-09-04
Check permissions at runtime


jupenur:
Currently, the WEBCAT makes no effort to verify that it actually has the permissions needed to function correctly. There are two scenarios that can lead to missing permissions:

 1. [Firefox treats MV3 host permissions as optional](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/host_permissions), allowing users to revoke them at any time.
 2. When used as a library, the host extension may not declare all the required permissions in its manifest.

WEBCAT should use the [permissions](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/permissions) API to check that it has the required permissions, notify the user if it does not, and provide steps to correct any issues.

PR #252, opened 2026-09-09
Check for required permissions at runtime

jupenur:
This is the first half of #247. A new `permissions` module allows declaring required permissions per-class, and logs a warning to the extension console when a permission is missing. This means library consumers are less likely to end up with insecure configurations due to simple omissions in the extension manifest.

image

The screenshot shows the warning after manually disabling `Access your data for all websites` in `about:addons`.

This PR does not surface permission warnings in the UI, which is the second half of #247. I'm envisioning a two-fold warning mechanism: firstly, a new warning icon for the page action, displayed on all pages regardless of enrollment status. And secondly, a page action popup that lists the missing permissions and provides guidance on resolving the issue. Implementing the UI requires adding persistence to the `ui` module, which is easiest to do if we refactor the module to a similar shape as the other stateful components of the extension. I'm going to work on that in a separate PR.


extension/src/webcat/updater.ts
```diff
@@ -94,6 +95,7 @@ export interface EnrollmentUpdater extends EventTarget {
  * Handles loading enrollment updates periodically from the endpoint,
  * storing them to the database, and notifying event consumers.
  */
+@permissions.require("alarms")
```
jupenur, line 98: I was debating whether to include `<all_urls>` as a required permission here. Strictly speaking the updater only needs permission to access `options.endpoint`, but it's not available before instantiation. I could add an explicit permission check inside the constructor if that makes sense?

extension/tests/browser/permissions.test.ts
```diff
@@ -0,0 +1,260 @@
...
+        "notARealPermission",
+        "scripting",
+        "https://example.org/*",
+      ]),
+    );
+    expect(permissions.getMissing(Class1)).toStrictEqual(
+      new Set(["tabs", "notARealPermission", "scripting"]),
+    );
+    expect(permissions.getMissing()).toStrictEqual(
+      new Set(["tabs", "notARealPermission", "scripting"]),
```
lsd-cat, line 181: It seems this test could fail in the real browser if `contains()` wasn't mocked, since in permissions.ts:183 rejections, which could happen for a non-existent permissions, are not caught.

extension/src/webcat/ui.ts
```diff
@@ -89,6 +90,7 @@ export function setErrorIcon(tabId: number) {
   });
 }
 
+permissions.require("tabs")(errorpage);
```
lsd-cat, line 93: Should it also need `permissions.require("webNavigation")(errorpage);` since below it could call `browser.webNavigation.getAllFrames`?

extension/src/webcat/response.ts
```diff
@@ -56,6 +57,10 @@ function assertHeadersAvailable<T>(
 /**
  * Validates responses from WEBCAT-enrolled origins.
  */
+@permissions.require("tabs")
```
lsd-cat, line 60: From [MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs) it seems `tabs` permissions might not be needed here, while probably `webRequestBlocking` is.

extension/src/browser/permissions.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  /**
+   * Checks the permissions of a given subject.
+   *
+   * @param subject The subject of the permissions.
+   * @returns A Promise that resolves to true if the subject has all the
+   *   permissions required.
+   */
+  async check(subject: Subject): Promise<boolean> {
+    const requirements = this.#requirements.get(subject) ?? new Set();
```
lsd-cat, line 165: Does this match subclasses too, or how do requirements propagate? e.g. does check(WebcatRequestHandler) include webRequest/webRequestBlocking from RequestHandler?

lsd-cat (approved):
As always this is great and the comments are minor!

extension/src/browser/permissions.ts
jupenur: There's no special handling for inheritance here; `check` needs to be called separately for each class in the inheritance chain. The `#handleAdded` and `#handleRemoved` methods already do that. Separate calls have the benefit of reporting permission errors more clearly and avoiding duplicated reporting.

extension/tests/browser/permissions.test.ts
jupenur: Yeah. The test is not trying to assert that permission checks can't reject; on the contrary, they should reject on a non-existent permission. The fake permission value is there to ensure `PermissionChecker` class works with arbitrary strings, so that if browsers introduce a new permission, nothing needs to be added or changed.

extension/src/webcat/response.ts
jupenur: Fixed in fe4f701.

extension/src/webcat/ui.ts
jupenur: Fixed in fe4f701.

extension/src/browser/permissions.ts
lsd-cat: Thanks for the explanation, got it now!

extension/tests/browser/permissions.test.ts
lsd-cat: Sorry, what I meant is that, even though it's not documented in MDN, I think the `browser.permission.contains` could throw for an unrecognized permission, but it doesn't in this test since it's mocked. It's possible that that can never happen anyway, so if that's the case it's oki for me as well!

jupenur: It does throw/reject, and I think that's what we want: if we accidentally write `@permissions.require("webRequestBlicking")` we'd want it to reject and log a big red error in the console instead of succeeding silently or complaining about a missing permission.

lsd-cat: sounds good!

PR #253, opened 2026-09-12
Refactor ui.ts

jupenur:
This PR refactors ui.ts to follow the same pattern as other parts of the codebase. That way we can more easily manage state as required for #247.

In addition to refactoring ui.ts, this PR contains minor interface changes to EnrollmentUpdater and HookBuilder to improve consistency.


extension/src/webcat.ts
lsd-cat: These should now both be `pagePaths` and `iconPaths`

extension/src/webcat/ui.ts
```diff
@@ -1,166 +1,150 @@
...
+    } else {
+      tabIds.add(details.tabId);
     }
-    await Promise.all(frameLookups);
-  } else {
-    tabIds.add(details.tabId);
-  }
 
-  const code = error?.code ?? "WEBCAT_ERROR_UNDEFINED";
+    const code = error?.code ?? "WEBCAT_ERROR_UNDEFINED";
```
lsd-cat: As error is now required, I think it van either be an optional parameter, or there's no need to fallback

lsd-cat (approved):
lgtm

Issue #246, opened 2026-09-03
Unused error codes


jupenur:
The codes `ERR_WEBCAT_BUNDLE_FETCH_PROMISE_MISSING` and `DEFAULT_FALLBACK_MISSING_FILE` are not currently used. The former should likely be removed, while the latter should be used [here](https://github.com/freedomofpress/webcat/blob/ffa7d79dd6b6a029de5bbc5425e010994b5a1da9/extension/src/webcat/validators.ts#L516) instead of `DEFAULT_FALLBACK_MISSING`.

`ERR_WEBCAT_CSP_PARSE_FAILED` is also unused. We do not parse CSP headers, so it seems like this one could also be removed.

lsd-cat:
Done in https://github.com/freedomofpress/webcat/pull/255

PR #255, opened 2026-09-16
Clean up unused error codes and return EXPIRED for stale Sigstore

### LINKED ISSUE #245: ERR_WEBCAT_MANIFEST_EXPIRED is only used for Sigsum manifests
The `ERR_WEBCAT_MANIFEST_EXPIRED` error code is only used here, for Sigsum manifests:

https://github.com/freedomofpress/webcat/blob/ffa7d79dd6b6a029de5bbc5425e010994b5a1da9/extension/src/webcat/validators.ts#L610-L616 

Sigstore manifests have an equivalent freshness check:

https://github.com/freedomofpress/webcat/blob/ffa7d79dd6b6a029de5bbc5425e010994b5a1da9/extension/src/webcat/validators.ts#L774-L776

Only, for Sigstore manifests, the more generic error code, `ERR_WEBCAT_MANIFEST_VERIFY_FAILED`, is used instead:

https://github.com/freedomofpress/webcat/blob/ffa7d79dd6b6a029de5bbc5425e010994b5a1da9/extension/src/webcat/validators.ts#L811-L813

Both manifest validation routines should return the `ERR_WEBCAT_MANIFEST_EXPIRED` error code when a manifest is expired.

lsd-cat:
Fixes https://github.com/freedomofpress/webcat/issues/245 and https://github.com/freedomofpress/webcat/issues/246


jupenur (approved):
LGTM

PR #256, opened 2026-09-17
Surface permission warnings in the UI

### LINKED ISSUE #247: Check permissions at runtime

jupenur:
Resolves #247. This PR adds new UI for permission warnings:

image

image

When a warning is triggered, the warning icon is shown in all tabs regardless of enrollment status. When permissions are corrected, the warning icon is removed from all tabs, but the green "ok" icons are not restored, as pages could have been corrupted while permissions were missing.

The icon for the warnings is currently the same as the one for errors; we can change that if we want to with the new icon designs.

The popup is currently only used for the warnings, but we can also populate it with information about enrolled websites at a later point. That information would be shown below the red warning box. The warning box can optionally include a link to a page explaining how to resolve the warning. That page remains to be created, but the plumbing to hook it up is there. E.g.:

image

The URL format of the popup page is like so:
```
moz-extension://3a3a9fbd-c94e-42b7-b590-a5842e2c4ed8/pages/popup.html#warnings={%22permissions%22:{%22message%22:%20%22This%20is%20an%20arbitrary%20warning%20with%20an%20attached%20URL.%22,%20%22url%22:%20%22https://example.com%22}}
```


lsd-cat:
I don't object to the patch per se, but I feel a bit skeptical of failing open, though I understand we might have no other choice. In the sense that, if the user revokes a permission that is fundamental for WEBCAT to provide its guarantees, then my thinking would be to either force them to enable what's required or remove the extension if they do not want it. But I guess that without the permissions themselves we cannot really block navigation or cause enough nuisance for that to be the outcome?

jupenur:
Yeah there's no way to really force the user. That's why I figured the best option would be to show the error icon on every page, so at least it doesn't go unnoticed. I also thought of adding nags via the notifications API, but that requires another permission too.

The thing we don't want to do is give the user an impression that the extension is operating correctly when it isn't. The permission toggles are buried kind of deep in the settings, so it's pretty difficult to notice otherwise if the extension doesn't notify you.

lsd-cat:
Thanks, I tested it a bit, maybe we can provide a button in the popup that directly triggers the granting request, using `browser.permissions.request({ origins: ["<all_urls>"] })`, so the user has a fast path at resolving the issue without navigating to settings.

I also tried to make it a bit more invasive, but I don't think it was an improvement:
 - Tried to display the error as a popup, but this opens a new window on my mac and seems too invasive and annoying.
 - Redirect enrolled tabs to the warning: this "works" but it doesn't stop execution, as the original tab start loading before the redirect. It might then be misleading to the user to think we blocked something we didn't?

Mostly the thing that isn't great is that in long-running sessions, in theory a single execution could steal cookies or secrets, and I'm not sure how to communicate to the user that something has happened, even if they now allow the extension again. Probably there's no good way around it though, it's also not a circumstance that we expect too often.

extension/src/webcat/ui.ts
```diff
@@ -144,7 +207,102 @@ export class WebcatUI extends BrowserChromeController {
...
+    }
+    this.#showWarnings = true;
+  }
+
+  /**
+   * Clears a specific warning from the UI.
+   *
+   * @param id The ID of the warning to clear.
+   */
+  async clearWarning(id: string) {
```
lsd-cat, line 244: Perhaps we should hide only on non-enrolled tabs and we should rather advise enrolled tabs to reload (or just do it?)

jupenur: I was thinking about that and decided against automatic reloading. There's a handful of reasons why I think it's not the best idea, but none of them are particularly strong:

 - We don't reload enrolled pages on fresh installation either, so this makes the behavior symmetric
 - Even if an app is compromised, the user might not want to lose the data entered into form fields, which would happen if we reload
 - *If* an app is compromised while permissions are unavailable, reloading would erase any visual evidence of it. By not reloading, a user who views a compromised tab can at least potentially see the data that they entered and act accordingly

That said, I kind of do like the idea of *not* hiding the icon and explicitly warning the user that the tab might have been compromised. But the warning would have to be different from the permission warning, and the state management would add some complexity. I think we should consider it, but maybe under a separate issue / in a new PR.

WDYT?

jupenur:
> Thanks, I tested it a bit, maybe we can provide a button in the popup that directly triggers the granting request, using `browser.permissions.request({ origins: ["<all_urls>"] })`, so the user has a fast path at resolving the issue without navigating to settings.

Good idea. Somehow that hadn't occurred to me. I'll look into it.

> I also tried to make it a bit more invasive, but I don't think it was an improvement:
> 
> * Tried to display the error as a popup, but this opens a new window on my mac and seems too invasive and annoying.
> * Redirect enrolled tabs to the warning: this "works" but it doesn't stop execution, as the original tab start loading before the redirect. It might then be misleading to the user to think we blocked something we didn't?

Agreed on both points.

> Mostly the thing that isn't great is that in long-running sessions, in theory a single execution could steal cookies or secrets, and I'm not sure how to communicate to the user that something has happened, even if they now allow the extension again. Probably there's no good way around it though, it's also not a circumstance that we expect too often.

Totally agreed. I think your suggestion in https://github.com/freedomofpress/webcat/pull/256#discussion_r4061982026 might help a bit, but there's no airtight solution here.

extension/src/webcat/ui.ts
lsd-cat: I do agree about automatic reloading, and sounds good in filing a separate issue and letting this go through!

jupenur:
Added a fix button in the warning and opened https://github.com/freedomofpress/webcat/issues/260 for the follow-up.

permissions

lsd-cat (approved):
lgtm!

PR #257, opened 2026-09-17
Fix race conditions in WASM verification

### LINKED ISSUE #235: Race condition in WASM hooks
In the WASM hooks, it seems that we use a mutable `ArrayBuffer` that coudls till be modified by anything holding the original reference, potentially after we have verified its hash but before it gets compiled. We should copy the buffer in a new one, and keep that as a referebce internally.

jupenur:
Fixes #235.


lsd-cat (approved):
lgtm, thanks!

Issue #241, opened 2026-08-26
Add lib-usage-based testing


lsd-cat:
https://github.com/freedomofpress/webcat/pull/218 introduced the usage of WEBCAT as a lib. We should add test to check we do not break any API-level functionality, probably something like a full suite on Firefox.

PR #258, opened 2026-09-21
Run the integration suite against a library-built extension

lsd-cat:
Implements #241. It just builds a basic extension, and run the same suite using it. It doesn't prove much on its own, besides that the basic API works, but it can be a starting point to add compatibility checks or regression tests as we work directly with other extensions. For instance, we could wait to merge https://github.com/freedomofpress/webcat/pull/178 beforehand, and then add a test featuring another listener.


test/libext/Makefile
```diff
@@ -0,0 +1,16 @@
+DIST := ../../dist
+OUT  := $(DIST)/webcat-lib-test.zip
+
+all:
+	$(MAKE) -C ../../extension package-lib-test
+	rm -rf node_modules/webcat
+	npm install --no-package-lock
```
jupenur: Ignoring lockfiles feels like inviting trouble. Can we split this into separate steps for installing the library without a lockfile and the rest of the dependencies with one?

jupenur (changes requested):
This is great. Just one request for the new Makefile.

test/libext/Makefile
lsd-cat: Done!

jupenur (approved):

Issue #237, opened 2026-08-25
Pre-compiled modules can be passed around same-site origins


lsd-cat:
Here's a regression test https://github.com/freedomofpress/webcat/compare/main...precompiled-module-poc

I remember discussing this @jupenur and agreeing it was either fixed or not-applicable, though the scope is pretty narrow since it has to be same-site.

jupenur:
I can't remember what we decided before, but here's what I'm thinking now:

1. There's no way to fix this in the WebAssembly hooks that doesn't break legitimate use cases. To prevent pre-compiled modules from being instantiated, we'd have to track the verified modules globally in `compile` and `compileStreaming`, and then reject in `instantiate` and `instantiateStreaming` if a module is not present in the global list of verified modules. But tracking would be limited to the current execution context, so passing a module to a same-origin worker and instantiating there would then no longer work. That's a legitimate thing that apps do.
2. We can't fix stupid, anyway. Yes, an app might pass a compiled module across origins using `postMessage` -- it might even do it safely, since instantiating and calling untrusted WebAssembly is safe as long as you don't give it any sensitive data or imports. But instantiating untrusted WASM and giving it access to the DOM is clearly reckless. On the other hand, `postMessage` itself is all kinds of dangerous, and if you want to be reckless, you could equally easily expose an RPC interface over `postMessage`; we're not stopping anyone from doing that either.
3. As the logical consequence of points 1 and 2, if we *really*, *really* wanted to fix this, the right point to address would probably be `postMessage` instead -- sanitizing messages that cross origins so that they cannot contain WebAssembly Modules. That's contrary to the test, which currently assumes that Modules *should* survive `postMessage` but not be allowed to instantiate. Proper sanitization would also be difficult and error prone due to the complexity of the structured clone algorithm. We'd have to check object properties, Map keys and values, Set members, and probably other things too that I can't think of off the top of my head; all of that recursively.

I vote we leave this as it is, and maybe explicitly document that `postMessage` is dangerous.

lsd-cat:
[...]

PR #261, opened 2026-09-22
Allow external iframes; always enforce isolation between non same-origin iframes

lsd-cat:
This forces `Origin-Agent-Cluster: ?1` on every enrolled document because same-site frames otherwise share the parent's agent cluster and can hand it compiled WebAssembly modules or merge origins via document.domain. See https://github.com/freedomofpress/webcat/issues/237

It blocks on explicit `?0`, values cause we would overwrite it anyway, and if the application relies on this behavior to work, it would fail anyway. This way it's pretty clear from the beginning that the feature is not available.

This also drops the `frame-src` and `child-src` restrictions. As explained in https://github.com/freedomofpress/webcat/issues/237#issuecomment-5776570113 there's some exclusions of when the policy is enforced: mostly when a tab or context has been loaded from the same origin but loaded without the header first. This can occur when a website enrolls while the user has a tab on the same origin already active.


jupenur:
I don't quite follow why OAC allows us to drop the framing restrictions? Wouldn't this now allow an enrolled origin to embed an unrelated non-enrolled origin and still get the green WEBCAT icon? Sure, the frame is isolated, but if it's rendered indistinguishably from same-origin content, that's still bad, right? A user interacting with a site would not be able to tell whether they're interacting with a WEBCAT-verified part or an untrusted part.

lsd-cat:
Sorry, I converged two things in one. Currently, we allow no iframes, unless they are enrolled. The main reason why I filed https://github.com/freedomofpress/webcat/issues/237 was that if we allowed untrusted iframes, the same-site ones could smuggle the WASM modules, but also otherwise gain access and control the parent. I think the OAC allows us to drop the overall restriction, without worrying about the privileges of same site-frames or treating them differently.

I don't think we can escape iframes as a concept for a bunch of use cases:
 - Sandboxing annoying content that won't ever be committed or open source: stuff like Google/Cloudflare captchas, but also video players or similar cases
 - Actually rendering untrusted content for the webapp own functionality: previews, code snippets in LLM chats

As in other cases, I don't think we can fully prevent bad things; we can't prevent a full page size iframe that is an untrusted website. But we also can't prevent the same happening via `srcdoc` or by fetching some random content and setting the entire `innerHTML` to it. So that boils down to the develoepr and the auditing process that can check whether the iframe is used properly.

In WAICT, I'd push for also enforcing some degree of sandboxing no matter what, to also ensure any child can't mess with the parent navigation. In WEBCAT, I don't think we can do that reliably (as it would require parsing the DOM, etc etc). I have to double check but I also understand that the general direction of WAICT is to just allow frames generally, and just apply the enforcement on those that are part of an enforced origin.

jupenur:
Thanks. Yeah. Makes sense. I'll take a thorough look at this PR when I have a couple of hours of uninterrupted focus time, probably tomorrow-ish, because I first want to properly understand OAC and what Firefox actually commits to.

lsd-cat:
Thank you! I'll also spend some more time on understanding the Firefox limitations, and possibly add documentation here or open a separate issue for it!

extension/src/webcat/handler.ts
```diff
@@ -264,6 +264,15 @@ export class WebcatRequestHandler extends RequestHandler {
       details.state.cachePartition,
     );
 
+    // Documents only (subresources ignore it). origin-keyed => same-site
+    // frames can't document.domain-merge into us or postMessage us a
+    // WebAssembly.Module. This is what replaces the old frame-src check.
+    if (details.state.isFrame) {
+      blockingResponse.set({
+        responseHeaders: withOriginAgentCluster(details.responseHeaders),
+      });
```
jupenur, line 273: Entirely optional, but it might be nicer if this was a method in the BlockingResponse class instead of a standalone function in the response module. I'm thinking something like this:

```ts
blockingResponse.setHeader("Origin-Agent-Cluster", "?1");
```

jupenur (approved):
> As explained in [#237 (comment)](https://github.com/freedomofpress/webcat/issues/237#issuecomment-5776570113) there's some exclusions of when the policy is enforced: mostly when a tab or context has been loaded from the same origin but loaded without the header first. This can occur when a website enrolls while the user has a tab on the same origin already active.

To close this gap, would it be possible to consult `window.originAgentCluster` in a content script and then somehow block if it's false? I guess blocking would be the tricky part; not sure how we'd achieve that in a content script. Maybe we should open a separate issue to investigate?

Other than that piece, this looks fine. Ok to merge from my perspective.

jupenur:
We should also update https://docs.webcat.tech/webapp-developers/CSP.html once this is in.

lsd-cat:
> To close this gap, would it be possible to consult `window.originAgentCluster` in a content script and then somehow block if it's false? I guess blocking would be the tricky part; not sure how we'd achieve that in a content script. Maybe we should open a separate issue to investigate?
> 

Thanks for the review! I was thinking a different approach, that would be subject to some performance measurements though. Should we rather scan existing tabs when a list update is shipped, and notify the user to close and reopen for protections to apply? I tested locally and indeed the case where the header is not initially set, and then it is set in the same tab for the origin is not covered. I can't think of others, so this should indeed only happen on list updates.

Edit: it probably makes sense to do the reverse: check for all the tabs in the list, which should be way faster.

jupenur:
Closing individual tabs would not be enough. The [historical agent cluster key map](https://html.spec.whatwg.org/multipage/document-sequences.html#historical-agent-cluster-key-map) is shared across a browsing context group, and the extension has no visibility into BCGs, so the user would have to relaunch the whole browser on every single update even if no enrolled tabs are open.

lsd-cat:
Yeah you're right, there's no escape from checking in-page :(

lsd-cat:
I've filed https://github.com/freedomofpress/webcat/issues/264 and updated the PR to have the class method. Docs PR will follow soon!

jupenur (approved):
Thanks! Looks great!

